PR #160, opened 2026-02-21
add local dev config for website

nikbpetrov:
will probably need better documentation and some additional dev/staging tooling (e.g. pull latest data from GCS instead of reusing old data from assets/)

### DIFF
--- a/src/www.forecastbench.org/Makefile
+++ b/src/www.forecastbench.org/Makefile
@@ -2,7 +2,7 @@ all :
 	$(MAKE) clean
 	$(MAKE) deploy
 
-.PHONY : all clean deploy
+.PHONY : all clean deploy dev-build dev-serve
 
 UPLOAD_DIR = upload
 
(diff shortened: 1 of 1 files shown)


houtanb (changes requested):

houtanb:
Also, re this, I want to serve local files, so do not want to pull data from GCS.

houtanb:
Yes, that's what happens. I'm responding to the comment [above](https://github.com/forecastingresearch/forecastbench/pull/160#issue-3971819458):
> will probably need better documentation and some additional dev/staging tooling (e.g. pull latest data from GCS instead of reusing old data from assets/)

Given this folder is for website development, the values of the leaderboard rarely matter. It's true that if we want to view the latest data generated locally on the website when we serve it locally, it needs to be copied over by hand but, thus far, this has not been frequently needed as an html leaderboard is also saved to the [datasets repo](https://github.com/forecastingresearch/forecastbench-datasets/tree/main/leaderboards/html).

houtanb (changes requested):

PR #176, opened 2026-04-20
fix: bug introduced in `7f6c3cda`

nikbpetrov:


houtanb:
@nikbpetrov there's another bug with this code.

Specifically, `_build_question_bank()` from `orchestration/_io.py` does `if "hash_mapping.json" not in filename` on line 164, but the change in 7f6c3cdac2add54443697756ce468c6091283abb stopped creating this file. Now, when a hash mapping is created, it's in `hash_mapping_<source>.json`

Please:
1. confirm that the hash mappings that are created are identical between the ones with `_source.json` and the original ones for all sources that create a hash mapping. These only exist for ACLED and Wikipedia I believe.
1. fix this and ~~add on top of this~~ create a new PR. What was the reasoning for changing this in the first place?

houtanb:
Merged in 02fda0c and ee93a00

PR #177, opened 2026-04-20
fix: bugs from `7f6c3cda`

nikbpetrov:
just setting `filename` to `hash_mapping.json` causes `gcp.storage.upload` to correctly upload the file to the remote

the second acled guard is a consequence of codex's investigation: the only reason resolve was not blowing up in my GCP project is due to the fact that both `{acled,wikipedia}/hash_mapping.json` and `{acled,wikipedia}/hash_mapping_{acled,wikipedia}.json` exist in the buckets, and are identical. If there was a hashing to be written, then the files would diverge - wikipedia would silently set resolution to `nan`, while acled would have blown up

### DIFF
--- a/src/orchestration/_io.py
+++ b/src/orchestration/_io.py
@@ -213,6 +213,7 @@ def upload_hash_mapping(raw_json: str, source_name: str) -> None:
         bucket_name=env.QUESTION_BANK_BUCKET,
         local_filename=local_filename,
         destination_folder=source_name,
+        filename="hash_mapping.json",
     )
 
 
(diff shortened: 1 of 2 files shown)


nikbpetrov:
Note that tests pass here. If you use the testsuite:
`fbtest test --pr 177 --only source.wikipedia.update --only source.acled.update`

you will notice that the current main branch code updates `{acled,wikipedia}/hash_mapping_{acled,wikipedia}.json` files, while the new code (from this PR) updates the correct `{acled,wikipedia}/hash_mapping.json` on the buckets

houtanb:
@codex review

houtanb:
@claude review

houtanb:
merged in 690904b6193f99c5d7d9c147ec2419ae38ccaddb and 1d576e77ff1cfebaeb089d03f7b7c78bfc1d858a

PR #192, opened 2026-05-16
init: manifold implementation

nikbpetrov:
In addition to the usual read/write patterns being updated during this refactor, note the updated bitwise OR update in `search_markets` - avoids having to pass `ids` around (and reduces, perhaps negligibly, memory footprint).

Note the second commit. I've found the fetch's 30s timeout was too quick during some testing in cases of high io in the GCP buckets, so bumped to 60s but I bumped further to match `max_time` from backoff.

Parity testing: old/new code matches perfectly when run together, with both fetch jobs finishing in ~10s, while for update jobs, the old code ran in 26mins vs 6mins for new code (see comment below). Both jobs' output differ ever so slightly compared to last night's prod data in my run (1 new id out of 2130, and 3 values slightly different out of 1981 values), all within expected deviations for a market source.

Full pipeline test: not done as per Slack discussion


nikbpetrov:
[dbb3f5d](https://github.com/forecastingresearch/forecastbench/pull/192/commits/dbb3f5ddf1198193e99ab778441db06bf2db5b9c) is new behaviour (tested with logs against old) but seems like quite the low-hanging fruit to save a ton of IO (8 mins of runtime for update job vs 28mins in ideal conditions for new code)

Edit: Actually, this was the intended behaviour behind the current prod code too ([here](https://github.com/forecastingresearch/forecastbench/blob/2594813d38d3e58f0f3a927c0016aa778b1745aa/src/questions/manifold/update_questions/main.py#L262-L270)), but there seems to be a bug: 
```python
# Regenerate resolution files in case they've been deleted
        resolved_files = gcp.storage.list_with_prefix(
            bucket_name=env.QUESTION_BANK_BUCKET, prefix=source
        )
        filename = f"{row['id']}.jsonl"
        if filename not in resolved_files:
            market = _get_market(row["id"])
            _create_resolution_file(dfq, index, market)
```
here, `resolved_files` has a prefix, while `filename` does not, resulting in a resolution file always being created and uploaded, and thus excessive writing to GCP.

houtanb:
@codex review

houtanb:
@claude review

nikbpetrov:
@houtanb I ran the full pipeline (I cancelled resolve and leaderboard after they started not to waste resources, but they start fine).

polymarket fetch job failed due to a timeout - seems to have been fetching stuff fine. Not sure what causes this at this point - I recall some issues with some providers?

Did not run acled as per previous agreement.

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+
+            # Build resolution file
+            existing_df = existing_resolution_files.get(row["id"])
+            df_res = self._build_resolution_file(
+                market=market,
+                market_info_resolution_datetime=dfq.at[index, "market_info_resolution_datetime"],
+                existing_df=existing_df,
+            )
+            if df_res is not None:
+                resolution_files[row["id"]] = df_res
```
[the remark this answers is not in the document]
houtanb: @nikbpetrov I'm confused here, can you help me understand.

Lines 162 and 166 are the exact same. So it would seem that the conditional `if df_res is not existing_df:` does _not_ impact whether or not the file is written because either way it's included in `resolution_files`. Can you explain this bit of code to me?

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+    @backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
+        max_time=200,
+        max_tries=10,
+        factor=2,
+        base=2,
+        on_backoff=data_utils.print_error_info_handler,
+    )
+    def _get_market(self, market_id: str) -> dict | None:
```
houtanb: never returns `None`

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+        # --- Update all unresolved questions ---
+        dfq["resolved"] = dfq["resolved"].astype(bool)
+        for index, row in dfq[~dfq["resolved"]].iterrows():
+            market = self._get_market(row["id"])
+            if market is None:
+                continue
+
+            # Assign market details to dfq row
+            dfq.at[index, "question"] = market["question"]
+            dfq.at[index, "background"] = market.get("textDescription", "")
```
houtanb: prefer `market["textDescription"]` so it errors out if the field doesn't exist

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+            dfq.at[index, "market_info_resolution_criteria"] = "N/A"
+            dfq.at[index, "market_info_open_datetime"] = dates.convert_epoch_time_in_ms_to_iso(
+                market["createdTime"]
+            )
+            dfq.at[index, "market_info_close_datetime"] = dates.convert_epoch_time_in_ms_to_iso(
+                market["closeTime"]
+            )
+            dfq.at[index, "url"] = market.get("url", "")
```
houtanb: prefer `market["url"]` so it errors out if the field doesn't exist

src/orchestration/func_manifold_update/requirements.txt
```diff
@@ -1,8 +1,10 @@
...
-pyarrow
-backoff
-certifi
 pandera
 termcolor
+requests
+certifi
+backoff
+numpy
+pyarrow
```
houtanb: can you check that `pyarrow` is used? Claude says it isn't

src/orchestration/func_manifold_fetch/requirements.txt
```diff
@@ -1,8 +1,10 @@
...
-pyarrow
-backoff
-certifi
 pandera
 termcolor
+requests
+certifi
+backoff
+numpy
+pyarrow
```
houtanb: can you check that `pyarrow` is used? Claude says it isn't

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+        else:
+            df = df.ffill()
+
+        df["id"] = market_id
+        return self._finalize_resolution_df(df)
+
+    @staticmethod
+    def _finalize_resolution_df(df: pd.DataFrame) -> DataFrame[ResolutionFrame]:
+        """Filter to benchmark period and validate as ResolutionFrame."""
+        df["date"] = pd.to_datetime(df["date"])
```
houtanb, line 411: this line modifies the df that was passed into the function.

Do `df = df.copy()` at the top to avoid changing df for the caller

src/sources/manifold.py
```diff
@@ -2,20 +2,401 @@
...
+    ) -> DataFrame[ResolutionFrame] | None:
+        """Build or update a resolution file for a single market."""
+        yesterday = dates.get_date_today() - timedelta(days=1)
+        market_id = market["id"]
+
+        # --- Already up-to-date check ---
+        if (
+            existing_df is not None
+            and not existing_df.empty
+            and pd.to_datetime(existing_df["date"].iloc[-1]).date() >= yesterday
```
houtanb: `existing_df["date"].max()` better than `existing_df["date"].iloc[-1]`?

nikbpetrov: yes!

src/sources/manifold.py
nikbpetrov: not 100% sure about this one:
1) definitely not a problem at the moment
2) maybe a problem in the future but that's a private method - shouldn't the caller/public method take care of `copy()`?
3) same pattern already used by infer (with no `copy()`)
4) still asking why? i guess there is some future requirement where a `copy()` makes sense but i am hesitant to build for future hypothetical requirements - especially at refactor stage

houtanb: OK, sounds good. I've become more weary of changing dataframes in functions as that can lead to surprising situations later on. But agreed that it's not necessary here at the moment

src/sources/manifold.py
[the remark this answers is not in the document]
nikbpetrov: of course, that's easy! I wrote the `if` check as a separate task, in a [separate commit](https://github.com/nikbpetrov/forecastbench/commit/dbb3f5ddf1198193e99ab778441db06bf2db5b9c), and forgot to delete the original `resolution_files` line.

src/orchestration/func_manifold_update/requirements.txt
nikbpetrov: my claude also brought this up but i still opted to include it as it was in the previous version of the code; testing now without it

nikbpetrov:
@houtanb Fixed issues and rebased on top of main. Tests pass and timing still heavily in favour of the new update func.

Still needs testing of full pipeline but as agreed will leave with you on the dev project.

src/sources/manifold.py
```diff
@@ -2,20 +2,400 @@
...
+        yesterday = dates.get_date_today() - timedelta(days=1)
+        market_id = market["id"]
+
+        # --- Already up-to-date check ---
+        if (
+            existing_df is not None
+            and not existing_df.empty
+            and pd.to_datetime(existing_df["date"].max()).date() >= yesterday
+        ):
+            return existing_df
```
houtanb: Take the example where this code runs twice in one day, e.g. the regular nightly run the manifold code ran correctly but something else broke so I manually rerun it in the morning. And, in the intervening period, a market resolves. Then you would have the case where `existing_df` is not `None`, is not empty, and the latest date is `yesterday` BUT this early return would prevent the last value of the resolution file from containing the resolution value (inserted below on line 366).

src/sources/manifold.py
```diff
@@ -2,20 +2,400 @@
...
+                market = self._get_market(row["id"])
+                if market is None:
+                    continue
+                df_res = self._build_resolution_file(
+                    market=market,
+                    market_info_resolution_datetime=row["market_info_resolution_datetime"],
+                    existing_df=None,
+                )
+                if df_res is not None:
+                    resolution_files[row["id"]] = df_res
```
houtanb, line 194: Might as well also set `dfq.at[index, "freeze_datetime_value"] = df_res["value"].iloc[-1]`. This might be a divergence from current code, so you could do this in a separate pull request, but probably best to update the freeze datetime value to accord with the resolved value. It doesn't make a practical difference, but is cleaner.

src/sources/manifold.py
```diff
@@ -2,20 +2,400 @@
...
+            df_new["resolved"] = False
+            df_new["freeze_datetime_value_explanation"] = "The market value."
+            df_new["market_info_resolution_datetime"] = "N/A"
+            dfq = pd.concat([dfq, df_new], ignore_index=True)
+
+        # --- Update all unresolved questions ---
+        dfq["resolved"] = dfq["resolved"].astype(bool)
+        for index, row in dfq[~dfq["resolved"]].iterrows():
+            market = self._get_market(row["id"])
+            if market is None:
```
houtanb: I don't think `market` be `None`. Maybe you need another check here?

src/sources/manifold.py
houtanb: Also, `df_res["value"].iloc[-1]` will work. But I wonder if we should get the value associated with the max date. Or, perhaps, part of loading a resolution file should be sorting it by date in which case this can remain as is.

src/sources/manifold.py
```diff
@@ -2,20 +2,400 @@
...
+        response = requests.get(endpoint, params=params, verify=certifi.where())
+        if not response.ok:
+            logger.error(
+                f"Request to endpoint failed for {endpoint}: {response.status_code} Error. "
+                f"{response.text}"
+            )
+            response.raise_for_status()
+
+        today = dates.get_date_today()
+        max_resolution_date = today + timedelta(days=_MAX_RESOLUTION_DATE_IN_DAYS)
```
houtanb: Lines 221-222: no need to recalculate on every call

houtanb (changes requested):
small fixes

src/sources/manifold.py
nikbpetrov: good catch; probably not; let this fail loudly -- `_get_market` should return an API response if response is OK -- tests catch for the expected API response; if that API changes, let this fail loudly and we resolve when that happens

src/sources/manifold.py
nikbpetrov: i guess, but what are we optimising for here?

speed? hardly a concern - for all 35 or so calls, that's in the micro-mili second range still

customizability? maybe, but as mentioned at some other place, we don't seem to have a requirement to customise the `max_resolution_date` (so that we choose an arg to `fetch()` or a class param), so I'd avoid building for future hypothetical needs, especially as part of a refactor. Correct me here if I am missing something.

readability? I'd argue not - it's exactly here, at `resolves_by`, that this `max_resolution_date` is used, so I'd rather not have to go through a few different files/places to figure out how this is calculated (especially given that this is not reused)

houtanb: Readability due to uniformity. There are other places where today is set but there are 1,000s of calls. Still not a big hit to time, but you'd likely pull it out at that point. So having it defined inside a function in some places and outside in others is inconsistent.

Idempotency. If you run the code at 23:59, with the code here, you'd have two different max_resolution_dates in a single run. 

This is not a big/important change, but I flagged it because I was flagging other more important things.

src/sources/manifold.py
nikbpetrov: good catch; fixed; conditioning not only on whether market `isResolved` but also on date to ensure all data is there (not 100% sure if that can fail, but seems more robust to me)

src/sources/manifold.py
nikbpetrov: this is done in the previous loop, [line 162](https://github.com/forecastingresearch/forecastbench/pull/192/changes/BASE..13f88b2b31a85d75b2d97f8601473c2ab2e279ff#diff-18061d7448c48b13fe7b925e90b62b157be737ae8fb92c25a533677d2c679b0cR162)

```python
if df_res is not None:
                dfq.at[index, "freeze_datetime_value"] = df_res["value"].iloc[-1]
                # if rebuilt, then write; else - skip
                if df_res is not existing_df:
                    logger.info(f"Rebuilt, will write - id={row['id']}")
                    resolution_files[row["id"]] = df_res
                else:
                    logger.info(f"Skipped writing to resolution files, not changed -id={row['id']}")
```

src/sources/manifold.py
nikbpetrov: okay, fair, opted to set this as an additional arg for `fetch()` and then pass onwards - not the cleanest but makes more sense than an additional class param (as this is not something that is used by `update()` or elsewhere) - also allows public customizability I guess

haven't tested this with the pipeline on GCP but it's only an interface change and `make test` should have caught any errors

nikbpetrov:
Tests pass again after the latest changes.

houtanb:
merged in 8965250

PR #195, opened 2026-05-29
init: metaculus implementation

nikbpetrov:
- timeout for fetch job increased to the same as infer; old job had 60s, but here we also have a max backoff of 300s; infer had 500s so there is definitely some inconsistency and I am not sure what rule to use; should suffice anyhow
- removed 3 stale deps: `pyarrow`, `tqdm`, `termcolor`
- implemented the same `today` kind of mechanism in `fetch()` as was done for [manifold](https://github.com/forecastingresearch/forecastbench/pull/192/changes/b698b422ce7b1d6f214a511377a9a9384b58e4ae)
- sticking with the 'fail loudly' principle: if one part of a single market update fails, it takes down the whole job with it
- (flagging just because one LLM surfaced this and I don't recall if we ever made this explicit, though I've been assuming it throughout): for sources upload happens in one go at the end, which is a divergence from the previous version of the code where process+upload happen on every loop iteration rather than process all resolutions files, collect, then return them (which now `update()` does) and then upload them in one go; I think this is good but just one to make it explicit once, if I haven't yet.
- note the edgecase fix in commit 3
- note the API divergence from commit 4: `existing_resolution_files` is not actually used here, so it will differ from other sources' `update()` interface, but I think that's totally fine

TESTS: All pass (old/new full parity, minor deviations from the prod code, all due do changes from the APIs). New update job is approx 2x faster (45mins for old, 23 mins for new). Fetch timing is the same.


src/orchestration/func_metaculus_fetch/Makefile
```diff
@@ -9,32 +9,38 @@ UPLOAD_DIR = upload
...
-Procfile:
-	cp -r $(ROOT_DIR)src/helpers/Procfile .
+Dockerfile: $(ROOT_DIR)src/helpers/Dockerfile.template
+	sed \
+		-e 's/REGION/$(CLOUD_DEPLOY_REGION)/g' \
+		-e 's/STACK/google-22-full/g' \
+		-e 's/PYTHON_VERSION/python312/g' \
+		$< > Dockerfile
 
-deploy : main.py .gcloudignore requirements.txt Procfile
```
houtanb, line 15: why do you remove `main.py` from the requirements for deployment?

src/orchestration/func_metaculus_fetch/Makefile
```diff
@@ -9,32 +9,38 @@ UPLOAD_DIR = upload
...
 	mkdir -p $(UPLOAD_DIR)/orchestration
 	cp $(ROOT_DIR)src/orchestration/__init__.py $(UPLOAD_DIR)/orchestration/
-	cp $(ROOT_DIR)src/orchestration/_io.py $(UPLOAD_DIR)/orchestration/
-	cp $^ $(UPLOAD_DIR)/
+	cp $(ROOT_DIR)src/orchestration/_source_io.py $(UPLOAD_DIR)/orchestration/
+	cp $(ROOT_DIR)src/_fb_types.py $(UPLOAD_DIR)/
+	cp $(ROOT_DIR)src/_schemas.py $(UPLOAD_DIR)/
+	cp main.py $(UPLOAD_DIR)/main.py
+	cp requirements.txt $(UPLOAD_DIR)/requirements.txt
+	cp Dockerfile $(UPLOAD_DIR)/
```
houtanb: once you add `main.py` back to the dependencies, you can replace lines 29-31 with `cp $^ $(UPLOAD_DIR)/`

src/orchestration/func_metaculus_fetch/Makefile
```diff
@@ -9,32 +9,38 @@ UPLOAD_DIR = upload
...
 	gcloud run jobs deploy \
 		func-data-metaculus-fetch \
 		--project $(CLOUD_PROJECT) \
 		--region $(CLOUD_DEPLOY_REGION) \
 		--tasks 1 \
 		--parallelism 1 \
-		--task-timeout 60s \
+		--task-timeout 540s \
```
houtanb, line 36: This takes longer now?

src/orchestration/func_metaculus_update/Makefile
```diff
@@ -9,20 +9,26 @@ UPLOAD_DIR = upload
...
-	cp -r $(ROOT_DIR)src/helpers/Procfile .
+Dockerfile: $(ROOT_DIR)src/helpers/Dockerfile.template
+	sed \
+		-e 's/REGION/$(CLOUD_DEPLOY_REGION)/g' \
+		-e 's/STACK/google-22-full/g' \
+		-e 's/PYTHON_VERSION/python312/g' \
+		$< > Dockerfile
 
-deploy : main.py .gcloudignore requirements.txt Procfile
+deploy : .gcloudignore requirements.txt Dockerfile
```
houtanb: same comment as above

src/orchestration/func_metaculus_update/Makefile
```diff
@@ -9,20 +9,26 @@ UPLOAD_DIR = upload
...
 	mkdir -p $(UPLOAD_DIR)/orchestration
 	cp $(ROOT_DIR)src/orchestration/__init__.py $(UPLOAD_DIR)/orchestration/
-	cp $(ROOT_DIR)src/orchestration/_io.py $(UPLOAD_DIR)/orchestration/
-	cp $^ $(UPLOAD_DIR)/
+	cp $(ROOT_DIR)src/orchestration/_source_io.py $(UPLOAD_DIR)/orchestration/
+	cp $(ROOT_DIR)src/_fb_types.py $(UPLOAD_DIR)/
+	cp $(ROOT_DIR)src/_schemas.py $(UPLOAD_DIR)/
+	cp main.py $(UPLOAD_DIR)/main.py
+	cp requirements.txt $(UPLOAD_DIR)/requirements.txt
+	cp Dockerfile $(UPLOAD_DIR)/
```
houtanb: same comment as above

houtanb (commented):
small improvements to makefiles

src/orchestration/func_metaculus_fetch/Makefile
nikbpetrov: explained in bullet 1 above

src/orchestration/func_metaculus_fetch/Makefile
nikbpetrov: copy-paste from infer/manifold

technically not a serious issue as things stand as `make metaculus-fetch` would fail on build if `main.py` is missing (as it is explicitly called as `cp main.py $(UPLOAD_DIR)/main.py`)

fix here or as a separate PR that touches infer/manifold too for consistency?

houtanb: fix here and on the polymarket PR. Make a separate PR for the makefiles that were already pushed that have that issue

src/orchestration/func_metaculus_fetch/Makefile
nikbpetrov: actually, that would be *quite* nice... currently we have:
```
.gcloudignore:
	cp -r $(ROOT_DIR)src/helpers/.gcloudignore .
```

but that's a fail as it copies stuff into the directory (e.g. `func_metaculus_fetch/.gcloudignore`) rather than the `upload` directory, where it's supposed to land

once again, a PR after this one to resolve these Makefile issues across the already refactored jobs?

none of these are dangerous, I think

houtanb:
merged in 8750ff387ea8b54a4e33f7206291a49c33b46c3f

PR #207, opened 2026-06-08
refactor: yfinance

nikbpetrov:
Found comments to be overly verbose on main so have stripped most. Code is pretty clear.

Date setting was pretty inconsistent - [78e608d](https://github.com/forecastingresearch/forecastbench/pull/207/commits/78e608d724ea90ab48265cdf8b5a60d25124b2e9) is my attempt at this - inconsistent w/ other sources that don't define an extra prop but instead pass args downstream, but that solution is UGLY here.


src/sources/yfinance.py
```diff
@@ -2,20 +2,446 @@
...
+        top_500 = self._get_sp500_tickers()
+        set_top_500 = set(top_500)
+        set_current = set(dfq["id"].unique()) if dfq is not None and "id" in dfq.columns else set()
+        all_tickers = list(set_top_500 | set_current)
+
+        logger.info(
+            f"Stock tickers not in top 500 but in current stocks: {set_current - set_top_500}"
+        )
+
+        current_time = dates.get_datetime_now()
```
nikbpetrov, line 74: note deviation: this used to be inside the loop

nikbpetrov:
Tests pass w/ perfect parity.

nikbpetrov:
Rebased on top of #215

houtanb:
Resolution fails with:
```
Traceback (most recent call last):
  File "/app/main.py", line 21, in <module>
    from sources.registry import SOURCES
  File "/app/sources/registry.py", line 18, in <module>
    from .yfinance import YfinanceSource
  File "/app/sources/yfinance.py", line 13, in <module>
    import yfinance as yf
ModuleNotFoundError: No module named 'yfinance'
```

PR #208, opened 2026-06-08
fix: strip duplicated resolution frame validation

### LINKED ISSUE #206: Strip duplicated resolution frame validation
Infer, for instance, explicitly [validates the resolution frame internally](https://github.com/forecastingresearch/forecastbench/blob/main/src/sources/infer.py#L396), thought that's unnecessary. These frames [get added](https://github.com/forecastingresearch/forecastbench/blob/e1cde0742deb50c25ecaf49d225d1a52b244bdb9/src/sources/infer.py#L155-L158) to the `resolution_frames` list that is returned as part of `UpdateResult` in `update()` and that `UpdateResult` [already defines that validation](https://github.com/forecastingresearch/forecastbench/blob/e1cde0742deb50c25ecaf49d225d1a52b244bdb9/src/_fb_types.py#L51-L70)

nikbpetrov:
fixes #206 

also happens to fix a dormant deviation for infer: original infer refactor [stripped infer's](https://github.com/forecastingresearch/forecastbench/commit/2594813d38d3e58f0f3a927c0016aa778b1745aa#diff-54049b12d5a7c1ea319b2d3d2d20287f3e4db72b3d19e1386870a1fa24328021L165) `astype(dtype=constants.RESOLUTION_FILE_COLUMN_DTYPE)` when building the resolution frame. Is now corrected and matches manifold/metaculus.

Let me know if I should run parity tests for all 3 sources - I will assume not.

### DIFF
--- a/src/sources/infer.py
+++ b/src/sources/infer.py
@@ -386,14 +386,14 @@ class InferSource(MarketSource):
 
     @staticmethod
     def _finalize_resolution_df(df: pd.DataFrame) -> DataFrame[ResolutionFrame]:
-        """Apply date filtering and return as validated ResolutionFrame.
+        """Apply date filtering and select resolution columns.
 
         Args:
             df (pd.DataFrame): Raw resolution data with id, date, value columns.
         """
         df["date"] = pd.to_datetime(df["date"])
         df = df[df["date"].dt.date >= constants.BENCHMARK_START_DATE_DATETIME_DATE]
-        return ResolutionFrame.validate(df[["id", "date", "value"]])
+        return df[["id", "date", "value"]].astype(dtype=constants.RESOLUTION_FILE_COLUMN_DTYPE)
 
     # ------------------------------------------------------------------
     # Private: question transformation
--- a/src/sources/manifold.py
+++ b/src/sources/manifold.py
@@ -407,11 +407,10 @@ class ManifoldSource(MarketSource):
 
     @staticmethod
     def _finalize_resolution_df(df: pd.DataFrame) -> DataFrame[ResolutionFrame]:
-        """Filter to benchmark period and validate as ResolutionFrame."""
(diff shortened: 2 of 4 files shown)


nikbpetrov:
@houtanb now also fixed polymarket as it was duplicating validation too but I'd also like to ask your advice on whether I should just refactor the semantics of resolution file building altogether - I don't like it anymore.

Some sources have both `_build_resolution_file` as well as `_finalize_resolution_df`, while others don't (seems like a data/market source split). I don't see the point in that - we don't use/centralize these abstractions. I'd suggest we stick to only `_build_resolution_df` that each source (optionally) has and all the stuff inside `_finalize_resolution_df` (for infer/manifold/metaculus) just gets moved to `_build_resolution_df` (which is the rename of `_build_resolution_file`.

What do you think?

src/sources/infer.py
```diff
@@ -386,14 +386,14 @@ def _build_resolution_file(
 
     @staticmethod
     def _finalize_resolution_df(df: pd.DataFrame) -> DataFrame[ResolutionFrame]:
```
houtanb, line 388: Should this return `pd.DataFrame` now that validation has moved to `UpdateResult.__post_init__`? Currently `DataFrame[ResolutionFrame]` implies this helper returns a validated schema.

houtanb: Same question for:
```python
    def _build_resolution_file(
        self,
        question: dict,
        resolved: bool,
        existing_df: DataFrame[ResolutionFrame] | None = None,
    ) -> DataFrame[ResolutionFrame]:
```

nikbpetrov: I want to keep the annotation for readability reasons. I appreciate the implicitness of the annotation but it's not decorated with `@pa.check_types` so validation does not apply.

nikbpetrov:
#215

PR #209, opened 2026-06-08
fix: rename metaculus `_create_resolution_frame` to `_build_resolution frame` and fix return annotation

nikbpetrov:
Not parity tested. Let me know if needed.


nikbpetrov:
#215

PR #211, opened 2026-06-09
feat: bucket set-up

RaidedCluster:
Adds Makefile, setup_permissions.sh, and variables.mk.example to automate IAM setup for the external submissions pipeline.

### DIFF
--- a/src/external-submissions/Makefile
+++ b/src/external-submissions/Makefile
@@ -1,0 +1,40 @@
+-include ../../variables.mk
+
+.PHONY: deploy-onboard setup-permissions clean
+
+REGION          := us-central1
+SA_EMAIL        ?= $(SUBMISSIONS_SA_EMAIL)
+ROOT_DIR        ?= ../../
+
+FUNC_DIR = functions
+
+ONBOARD_ENV_VARS = \
+	RUN_MODE=$(RUN_MODE),\
+	SMTP_USER=$(SMTP_USER),\
+	SMTP_PASSWORD=$(SMTP_PASSWORD),\
+	UPLOAD_BUCKET=$(SUBMISSIONS_BUCKET),\
+	NEXT_DUE_DATE=$(NEXT_DUE_DATE)
+
+deploy-onboard:
+	cp $(ROOT_DIR)src/helpers/email.py $(FUNC_DIR)/onboard/email_utils.py
+	gcloud functions deploy onboard-team \
+		--gen2 \
+		--project=$(CLOUD_PROJECT) \
+		--region=$(REGION) \
+		--runtime=python312 \
+		--source=$(FUNC_DIR)/onboard \
+		--entry-point=onboard \
+		--trigger-http \
+		--no-allow-unauthenticated \
+		--service-account=$(SA_EMAIL) \
+		--memory=256Mi \
+		--timeout=120s \
+		--set-env-vars=$(ONBOARD_ENV_VARS)
+	rm -f $(FUNC_DIR)/onboard/email_utils.py
+
+setup-permissions:
+	eval $$(cat ../../variables.mk | grep -v '^#' | xargs) \
+		python $(FUNC_DIR)/setup_permissions.py
+
+clean:
+	rm -f $(FUNC_DIR)/onboard/email_utils.py
--- a/src/external-submissions/PIPELINE.md
+++ b/src/external-submissions/PIPELINE.md
@@ -1,0 +1,93 @@
+# ForecastBench External Submission Pipeline
+
+## Firestore — `teams` collection
+
+One document per team.
+
+| Field | Notes |
+| --- | --- |
+| `team_id` | `team1`, `team2`, ... — permanent internal ID, used as GCS folder name |
+| `team_name` | Optional internal label (unique). Used to distinguish multiple teams from the same org (e.g. "GDM A", "GDM B"). Never shown publicly or in emails. |
+| `organization` | Public name. `"Anonymous N"` if anonymous. |
+| `deanonymized_organization` | Always the real org name. Never shown publicly. |
+| `emails` | Used for IAM and email notifications |
+| `service_accounts` | GCP service accounts for automated uploads. No emails sent to these. |
+| `anonymous` | bool |
+| `created_at` | Firestore server timestamp |
+| `active` | bool — set to false on removal |
+
+### Counter document
+
+`counters/teams` holds `{team_count: N, anon_count: M}` for atomic ID allocation.
+Initialize before first deployment (set N and M to the current team and anon counts):
+
+```python
+db.collection("counters").document("teams").set({"team_count": N, "anon_count": M})
+```
+
+---
+
+## Register a new team
+
+POST to the `onboard-team` Cloud Function:
+
+```json
+{
+  "organization": "Acme Corp",
+  "team_name": "acme-a",
+  "emails": ["alice@acme.com", "bob@acme.com"],
+  "service_accounts": ["submissions@acme.iam.gserviceaccount.com"],
+  "anonymous": false
+}
+```
+
+Fields:
+
+- `organization` (required) — real org name
+- `team_name` (optional) — internal label, must be unique
+- `emails` (required) — list of member addresses; must be Gmail/Google Workspace for GCS access
+- `service_accounts` (optional) — GCP SAs; always get GCS access
+- `anonymous` (optional, default false) — if true, public name becomes `"Anonymous N"`
+
+The function:
+
+1. Allocates the next `teamN` ID atomically via `counters/teams`
+2. Creates a `gs://<bucket>/teamN/.keep` placeholder
+3. Grants `roles/storage.objectUser` + `roles/storage.objectViewer` on the `teamN/` prefix
+4. Writes the Firestore document
+5. Sends a welcome email to `emails`
+
+If any email is not a Google account, registration succeeds but a warning is returned — those members won't be able to upload to GCS directly.
+
+---
+
+## Remove a team
+
+DELETE to the `onboard-team` Cloud Function:
+
+```json
+{ "team_id": "team7" }
+```
+
+Revokes GCS access and marks the team inactive. IAM removal failure returns a 500 — the team is **not** deactivated if permissions cannot be revoked.
+
+---
+
+## Deploy
+
+From `src/external-submissions/`:
+
+```bash
+make deploy-onboard
+```
+
+Required variables in `variables.mk` (at repo root):
+
+```makefile
+CLOUD_PROJECT=...
+SUBMISSIONS_SA_EMAIL=...
+SUBMISSIONS_BUCKET=...
+SMTP_USER=...
+SMTP_PASSWORD=...
+NEXT_DUE_DATE=YYYY-MM-DD
+```
--- a/src/external-submissions/functions/onboard/main.py
+++ b/src/external-submissions/functions/onboard/main.py
@@ -1,0 +1,293 @@
+"""ForecastBench team onboarding handler.
+
+Each team gets a unique internal ID (team1, team2, ...) used as their GCS folder name.
+One organization can have multiple teams — team_name is an optional internal label to
+distinguish them (e.g. "GDM A", "GDM B"). It is never shown in emails.
+
+Accepts JSON POST with:
+  - organization: real org name (required, stored privately if anonymous)
+  - team_name: optional internal label (must be unique if provided)
+  - emails: list of member email addresses (required)
+  - service_accounts: list of GCP service accounts (optional, no emails sent)
+  - anonymous: bool (optional, default false)
+
+Environment variables:
+  - UPLOAD_BUCKET: GCS bucket name
+  - NEXT_DUE_DATE: next forecast due date (YYYY-MM-DD), included in welcome email
+"""
+
+import json
+import os
+import traceback
+
+import dns.resolver
+from email_utils import send_welcome
+from google.cloud import firestore, storage
+
+UPLOAD_BUCKET = os.environ.get("UPLOAD_BUCKET", "")
+NEXT_DUE_DATE = os.environ.get("NEXT_DUE_DATE", "")
+
+db = firestore.Client()
+gcs = storage.Client()
+
+
+@firestore.transactional
+def _allocate_ids(transaction, counter_ref, anonymous):
+    """Atomically allocate the next team_id and (if anonymous) anon number."""
+    snap = counter_ref.get(transaction=transaction)
+    data = snap.to_dict() if snap.exists else {"team_count": 0, "anon_count": 0}
+    team_n = data.get("team_count", 0) + 1
+    anon_n = data.get("anon_count", 0) + (1 if anonymous else 0)
+    transaction.set(counter_ref, {"team_count": team_n, "anon_count": anon_n})
+    return f"team{team_n}", anon_n
+
+
+def _team_name_taken(db, team_name):
+    """Return True if any active team already has this internal label."""
+    results = list(
+        db.collection("teams")
+        .where("team_name", "==", team_name)
+        .where("active", "==", True)
+        .stream()
+    )
+    return len(results) > 0
+
+
+GOOGLE_MX_SUFFIXES = ("google.com", "googlemail.com")
+
+
+def _is_google_account(email):
+    """Return True if the email is a Gmail or Google Workspace account."""
+    if email.endswith("@gmail.com"):
+        return True
+    domain = email.split("@")[-1]
+    try:
+        records = dns.resolver.resolve(domain, "MX")
+        return any(
+            str(r.exchange).rstrip(".").endswith(suffix)
+            for r in records
+            for suffix in GOOGLE_MX_SUFFIXES
+        )
+    except Exception:
+        return False
+
+
+def _warn_non_google_emails(emails):
+    """Return a warning string if any emails are not Google accounts, else None."""
+    non_google = [e for e in emails if not _is_google_account(e)]
+    if non_google:
+        return (
+            f"{len(non_google)} email(s) do not appear to be Google accounts: {non_google}. "
+            f"GCS upload permissions require Gmail or Google Workspace accounts. "
+            f"See wiki for handling non-Google accounts."
+        )
+    return None
+
+
+def _set_folder_permissions(bucket_name, team_id, emails, service_accounts):
+    """Grant objectUser + objectViewer on the team's folder prefix only."""
+    client = storage.Client()
+    bucket = client.bucket(bucket_name)
+    policy = bucket.get_iam_policy(requested_policy_version=3)
+    policy.version = 3
+
+    folder_prefix = f"projects/_/buckets/{bucket_name}/objects/{team_id}/"
+
+    def _principal(email):
+        if email.endswith(".iam.gserviceaccount.com") or "gserviceaccount" in email:
+            return f"serviceAccount:{email}"
+        return f"user:{email}"
+
+    members = set(_principal(e) for e in emails + service_accounts)
+    condition = {
+        "title": f"{team_id} folder access",
+        "expression": f'resource.name.startsWith("{folder_prefix}")',
+    }
+    for role in ("roles/storage.objectUser", "roles/storage.objectViewer"):
+        policy.bindings.append({"role": role, "members": members, "condition": condition})
+
+    bucket.set_iam_policy(policy)
+
+
+def _remove_folder_permissions(bucket_name, team_id, emails, service_accounts):
+    """Remove all IAM bindings for a team's folder from the bucket."""
+    client = storage.Client()
+    bucket = client.bucket(bucket_name)
+    policy = bucket.get_iam_policy(requested_policy_version=3)
+    policy.version = 3
+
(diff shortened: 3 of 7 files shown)


src/helpers/utils.py
```diff
@@ -0,0 +1,39 @@
+"""Submission pipeline date utilities.
```
houtanb: rename this file `submissions.py`

src/helpers/email.py
```diff
@@ -0,0 +1,84 @@
...
+   gcloud storage cp your-file.json gs://{upload_bucket}/{team_id}/
+
+To test your upload permissions before the due date:
+   gcloud storage cp test.json gs://{upload_bucket}/{team_id}/test/
+
+Deadline: 23:59:59 UTC on the forecast due date.
+Max submissions: 3 per round (one per model).
+
+Validate your file before submitting:
+{VALIDATE_URL}
```
houtanb: validation is not here yet, so delete these lines

src/helpers/email.py
```diff
@@ -0,0 +1,84 @@
...
+        anonymous (bool): Whether the team is anonymous.
+    """
+    subject = "ForecastBench — Your team has been registered"
+    body = f"""Hi,
+
+Your team has been registered on ForecastBench.
+
+Team: {display_name}
+Upload folder: gs://{upload_bucket}/{team_id}/
+
```
houtanb: mention the next forecast due date

src/helpers/utils.py
```diff
@@ -0,0 +1,39 @@
...
+Uses src/helpers/dates.py (copied into each function at deploy time).
+Adds mock-date support for dev/test environments.
+
+Set BUILD_ENV=dev and MOCK_DATE=YYYY-MM-DD to simulate a different current date.
+"""
+
+import os
+from datetime import datetime, timezone
+
+BUILD_ENV = os.environ.get("BUILD_ENV", "prod")
```
houtanb: use `RunMode` instead of this. See https://github.com/forecastingresearch/forecastbench/blob/main/src/helpers/constants.py#L102-L119

src/external-submissions/PIPELINE.md
```diff
@@ -0,0 +1,64 @@
...
+## Firestore — `teams` collection
+
+One document per team.
+
+| Field | Notes |
+|---|---|
+| `team_id` | `team1`, `team2`, ... — permanent internal ID, used as GCS folder name |
+| `team_name` | Optional display name (unique). Shown in emails. |
+| `organization` | Public name. `"Anonymous N"` if anonymous. |
+| `original_organization` | Always the real org name. Never shown publicly. |
```
houtanb: Maybe call `deanonymized_organization`

src/external-submissions/PIPELINE.md
```diff
@@ -0,0 +1,64 @@
+# ForecastBench External Submission Pipeline
+
+## Firestore — `teams` collection
+
+One document per team.
+
+| Field | Notes |
+|---|---|
+| `team_id` | `team1`, `team2`, ... — permanent internal ID, used as GCS folder name |
+| `team_name` | Optional display name (unique). Shown in emails. |
```
houtanb: This should not be shown in emails. It's only internal to differentiate between the 3 GDM teams

src/external-submissions/PIPELINE.md
```diff
@@ -0,0 +1,64 @@
...
+| `organization` | Public name. `"Anonymous N"` if anonymous. |
+| `original_organization` | Always the real org name. Never shown publicly. |
+| `emails` | Used for IAM and email notifications |
+| `service_accounts` | GCP service accounts for automated uploads. No emails sent to these. |
+| `anonymous` | bool |
+| `rounds_participated` | Round dates with at least one valid submission |
+| `models_used` | Populated from submitted forecast files, not from registration |
```
houtanb: Ok to delete I think. Did you need this for something in another commit? If so, delete for now then add back later

src/external-submissions/Makefile
```diff
@@ -0,0 +1,104 @@
...
+		--memory=256Mi \
+		--timeout=$4 \
+		--min-instances=1 \
+		--set-env-vars=$(SUBMISSIONS_ENV_VARS)
+	$(call CLEAN_COMMON,$2)
+endef
+
+deploy: deploy-validate deploy-onboard deploy-upload deploy-post-round deploy-send-reminders
+
+deploy-validate:
```
houtanb: remove the rules that don't belong in this commit. mentioning once here but applies to other rules too

src/external-submissions/Makefile
```diff
@@ -0,0 +1,104 @@
...
+	      $(FUNC_DIR)/$1/email_utils.py $(FUNC_DIR)/$1/utils.py
+endef
+
+define DEPLOY_HTTP_FUNC
+	$(call COPY_COMMON,$2)
+	gcloud functions deploy $1 \
+		--gen2 \
+		--project=$(CLOUD_PROJECT) \
+		--region=$(REGION) \
+		--runtime=python311 \
```
houtanb: Why Python 3.11?

src/external-submissions/Makefile
```diff
@@ -0,0 +1,104 @@
...
+	UPLOAD_BUCKET=$(SUBMISSIONS_BUCKET),\
+	INTERSTITIAL_BUCKET=$(SUBMISSIONS_INTERSTITIAL_BUCKET),\
+	HISTORY_BUCKET=$(SUBMISSIONS_HISTORY_BUCKET),\
+	FORECAST_SETS_BUCKET=$(FORECAST_SETS_BUCKET),\
+	VALIDATE_URL=$(VALIDATE_URL)
+
+FUNC_DIR = functions
+
+define COPY_COMMON
+	cp $(ROOT_DIR)src/helpers/submission_validation.py $(FUNC_DIR)/$1/validation.py
```
houtanb: file doesn't exist, rule broken. 

I'm going to stop reviewing the Makefile now. Ensure you can deploy this function and it runs on GCP and does what you want then submit an updated PR

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+
+SUBMISSIONS_SA_EMAIL=submissions@<project>.iam.gserviceaccount.com
+SUBMISSIONS_BUCKET=<project>-submissions
+SUBMISSIONS_INTERSTITIAL_BUCKET=<project>-submissions-interstitial
+SUBMISSIONS_HISTORY_BUCKET=<project>-submissions-history
+SUBMISSIONS_DEPLOYER=<your-email>@forecastingresearch.org
+
+SMTP_USER=
+SMTP_PASSWORD=
+FRI_EMAIL=forecastbench@forecastingresearch.org
```
houtanb: This comes from `src/constants.py`

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+# CLOUD_PROJECT and FORECAST_SETS_BUCKET come from the root variables.mk.
+# Add the submissions-specific vars below to your root variables.mk.
+
+SUBMISSIONS_SA_EMAIL=submissions@<project>.iam.gserviceaccount.com
+SUBMISSIONS_BUCKET=<project>-submissions
+SUBMISSIONS_INTERSTITIAL_BUCKET=<project>-submissions-interstitial
+SUBMISSIONS_HISTORY_BUCKET=<project>-submissions-history
+SUBMISSIONS_DEPLOYER=<your-email>@forecastingresearch.org
+
+SMTP_USER=
```
houtanb: should be a GCP secret

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+SMTP_PASSWORD=
```
houtanb: should be a GCP secret

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+FRI_EMAIL=forecastbench@forecastingresearch.org
+VALIDATE_URL=
```
houtanb: delete

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+
+BUILD_ENV=prod
```
houtanb: use RunMode

src/external-submissions/functions/variables.mk.example
```diff
@@ -0,0 +1,18 @@
...
+# variables.mk is gitignored — never commit real values.
+#
+# CLOUD_PROJECT and FORECAST_SETS_BUCKET come from the root variables.mk.
+# Add the submissions-specific vars below to your root variables.mk.
+
+SUBMISSIONS_SA_EMAIL=submissions@<project>.iam.gserviceaccount.com
+SUBMISSIONS_BUCKET=<project>-submissions
+SUBMISSIONS_INTERSTITIAL_BUCKET=<project>-submissions-interstitial
+SUBMISSIONS_HISTORY_BUCKET=<project>-submissions-history
+SUBMISSIONS_DEPLOYER=<your-email>@forecastingresearch.org
```
houtanb: put all of this in `variables.mk` in the root dir and delete this file

src/external-submissions/functions/setup_permissions.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
houtanb: Instead of a permissions script, can you do this in Python using `google-cloud-resource-manager`?

Codex says something like:

```python
from google.cloud import resourcemanager_v3
from google.iam.v1 import iam_policy_pb2, policy_pb2


def grant_role_on_folder(
    folder_id: str,
    member: str,
    role: str,
) -> None:
    """
    Example:
      folder_id = "123456789012"
      member = "serviceAccount:my-sa@my-project.iam.gserviceaccount.com"
      role = "roles/viewer"
    """
    client = resourcemanager_v3.FoldersClient()
    resource = f"folders/{folder_id}"

    policy = client.get_iam_policy(
        request=iam_policy_pb2.GetIamPolicyRequest(resource=resource)
    )

    for binding in policy.bindings:
        if binding.role == role:
            if member not in binding.members:
                binding.members.append(member)
            break
    else:
        policy.bindings.append(
            policy_pb2.Binding(
                role=role,
                members=[member],
            )
        )

    client.set_iam_policy(
        request=iam_policy_pb2.SetIamPolicyRequest(
            resource=resource,
            policy=policy,
        )
    )
```

src/external-submissions/functions/deploy.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
```
houtanb: I'm a bit lost as to why there's a `deploy.sh` script when you have a Makefile

src/external-submissions/functions/onboard/main.py
```diff
@@ -0,0 +1,285 @@
+"""ForecastBench team onboarding handler.
+
+Each team gets a unique internal ID (team1, team2, ...) used as their GCS folder name.
+One organization can have multiple teams (e.g. GDM has team1 and team2).
+Teams can optionally provide a display name to distinguish multiple teams from the same org.
```
houtanb: > Teams can optionally provide a display name to distinguish multiple teams from the same org.

This is not correct

houtanb (changes requested):

houtanb:
The review above was mine. This is from Codex. I haven't looked into the details of it but skimmed it and it seems useful
```
  1. Critical: deploy paths are broken. make deploy-onboard copies missing files before deploy: src/helpers/
     submission_validation.py, functions/common/email.py, and functions/common/utils.py do not exist
     (forecastbench/src/external-submissions/Makefile:26). The documented ./src/external-submissions/
     functions/deploy.sh onboard path is also broken because ROOT_DIR=../../ is evaluated after the script cds
     into functions/, so helper copies point at the wrong directory (forecastbench/src/external-submissions/
     functions/deploy.sh:18, forecastbench/src/external-submissions/functions/deploy.sh:110). The script is
     also committed 100644, but docs invoke it directly (forecastbench/src/external-submissions/
     PIPELINE.md:63).

  2. Critical: setup-permissions likely fails. The Makefile passes PROCESSING_BUCKET=$(FORECAST_SETS_BUCKET),
     but the script requires FORECAST_SETS_BUCKET (forecastbench/src/external-submissions/Makefile:92,
     forecastbench/src/external-submissions/functions/setup_permissions.sh:13). My make -n setup-permissions
     confirmed the required variable is not passed.

  3. High: remove can report access revoked while access remains. _handle_remove catches IAM removal errors,
     then still marks the team inactive and returns "GCS access revoked" (forecastbench/src/external-
     submissions/functions/onboard/main.py:170). That is a security failure mode: deleted teams can retain
     bucket access.

  4. High: onboard can report success without granting upload access. _iam() catches and only prints IAM setup
     failures, while the API still returns 201 (forecastbench/src/external-submissions/functions/onboard/
     main.py:240). Since setting conditional IAM also requires uniform bucket-level access【turn1view0†L, any
     bucket missing that precondition silently creates a team that cannot upload.

  5. High: team IDs and anonymous names race. _next_team_id() and _next_anon_number() count existing Firestore
     docs outside a transaction (forecastbench/src/external-submissions/functions/onboard/main.py:34,
     forecastbench/src/external-submissions/functions/onboard/main.py:40). Concurrent requests can assign the
     same teamN and same GCS prefix.

  6. High: the promised folder workflow does not match GCS IAM semantics. The condition only grants object-
     prefix access (forecastbench/src/external-submissions/functions/onboard/main.py:103), but Cloud Storage
     says storage.objects.list is bucket-level and cannot be prefix-restricted with resource.name. The code/
     email tell users to use GCP Console and gcloud storage cp (forecastbench/src/external-submissions/
     functions/onboard/main.py:266, forecastbench/src/helpers/email.py:63), but Google documents list
     permissions as required for CLI/Console upload verification.

  7. High: non-Google email handling contradicts the docs. Docs say registration succeeds with a warning for
     non-Google accounts (forecastbench/src/external-submissions/PIPELINE.md:47), but IAM setup still includes
     those principals in the same binding (forecastbench/src/external-submissions/functions/onboard/
     main.py:238). If one invalid principal makes set_iam_policy fail, no one gets access, and the API still
     returns success.

  8. Medium: external teams get delete/overwrite powers. The code grants roles/storage.objectUser
     (forecastbench/src/external-submissions/functions/onboard/main.py:107), which includes delete/update
     permissions. Google notes delete is only needed for overwrites. If submissions should be auditable, use
     objectCreator/custom roles or object versioning/retention.

  9. Medium: side effects are not atomic. The handler creates .keep, writes Firestore, then mutates IAM and
     sends email (forecastbench/src/external-submissions/functions/onboard/main.py:220). Any partial failure
     leaves orphan folders, active teams without access, or emails sent before permissions are valid.

  10. Medium: secrets are deployed as plain env vars. SMTP_PASSWORD is passed through --set-env-vars
     (forecastbench/src/external-submissions/Makefile:13, forecastbench/src/external-submissions/functions/
     deploy.sh:43). Google recommends Secret Manager for sensitive Cloud Run/Functions config and warns env-
     var secrets can leak through misconfiguration/logging.

  11. Medium: shell scripts continue after failures. Neither script uses set -euo pipefail; failed cp, gcloud,
     or rm commands can be followed by "deployed" / "Done" messages (forecastbench/src/external-submissions/
     functions/deploy.sh:60, forecastbench/src/external-submissions/functions/setup_permissions.sh:22).

  12. Low/medium: docs and filename contract disagree. API instructions say {round_date}.{organization}.
     {N}.json (forecastbench/src/external-submissions/functions/onboard/main.py:269); email says
     {forecast_due_date}.{organization}.{N}.json (forecastbench/src/helpers/email.py:62). Existing code uses
     forecast_due_date.

  13. Low: CORS preflight omits Authorization. Browser calls with bearer tokens will fail preflight because
     only Content-Type is allowed (forecastbench/src/external-submissions/functions/onboard/main.py:140).

  14. Low: no tests cover this. There are no tests for team ID allocation, IAM failure behavior, removal,
     validation, or deploy preflight. This code is mostly cloud side effects, so fake Firestore/GCS tests
     would catch the biggest regressions.
```

src/external-submissions/Makefile
```diff
@@ -0,0 +1,40 @@
...
+SA_EMAIL        ?= $(SUBMISSIONS_SA_EMAIL)
+ROOT_DIR        ?= ../../
+
+FUNC_DIR = functions
+
+ONBOARD_ENV_VARS = \
+	RUN_MODE=$(RUN_MODE),\
+	SMTP_USER=$(SMTP_USER),\
+	SMTP_PASSWORD=$(SMTP_PASSWORD),\
+	UPLOAD_BUCKET=$(SUBMISSIONS_BUCKET),\
```
houtanb, line 15: `UPLOAD_BUCKET` should be called `SUBMISSIONS_BUCKET` as the first one is not informative.

also, you need to modify `variables.example.mk` to include any variables that need to be set

src/external-submissions/functions/setup_permissions.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
+"""Set up IAM permissions for the ForecastBench submissions service account.
+
+Run once per environment after creating the service account:
```
houtanb, line 4: what does "environment" mean here?

src/external-submissions/functions/setup_permissions.py
```diff
@@ -0,0 +1,129 @@
...
+
+    eval $(cat ../../variables.mk | grep -v '^#' | xargs) python setup_permissions.py
```
houtanb, line 6: remove `grep -v '^#' | `

src/external-submissions/functions/setup_permissions.py
```diff
@@ -0,0 +1,129 @@
...
+
+Requirements:
+    pip install google-cloud-storage google-cloud-resource-manager
```
houtanb, line 9: remove these two lines

src/external-submissions/functions/setup_permissions.py
houtanb: you can remove this comment altogether

src/external-submissions/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	NEXT_DUE_DATE=$(NEXT_DUE_DATE)
+
+deploy-onboard:
```
houtanb, line 18: the files this deplooyment depends on should be on this line

src/external-submissions/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	cp $(ROOT_DIR)src/helpers/email.py $(FUNC_DIR)/onboard/email_utils.py
```
houtanb, line 19: why is this file renamed in the copy?

src/external-submissions/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	gcloud functions deploy onboard-team \
+		--gen2 \
+		--project=$(CLOUD_PROJECT) \
+		--region=$(REGION) \
+		--runtime=python312 \
```
houtanb, line 24: why python 3.12?

src/external-submissions/Makefile
```diff
@@ -0,0 +1,40 @@
...
+		--entry-point=onboard \
+		--trigger-http \
+		--no-allow-unauthenticated \
+		--service-account=$(SA_EMAIL) \
+		--memory=256Mi \
+		--timeout=120s \
+		--set-env-vars=$(ONBOARD_ENV_VARS)
+	rm -f $(FUNC_DIR)/onboard/email_utils.py
+
+setup-permissions:
```
houtanb, line 35: please remove this rule

src/external-submissions/PIPELINE.md
```diff
@@ -0,0 +1,93 @@
...
+| --- | --- |
+| `team_id` | `team1`, `team2`, ... — permanent internal ID, used as GCS folder name |
+| `team_name` | Optional internal label (unique). Used to distinguish multiple teams from the same org (e.g. "GDM A", "GDM B"). Never shown publicly or in emails. |
+| `organization` | Public name. `"Anonymous N"` if anonymous. |
+| `deanonymized_organization` | Always the real org name. Never shown publicly. |
+| `emails` | Used for IAM and email notifications |
+| `service_accounts` | GCP service accounts for automated uploads. No emails sent to these. |
+| `anonymous` | bool |
+| `created_at` | Firestore server timestamp |
+| `active` | bool — set to false on removal |
```
houtanb, line 17: in the above list add two more columns showing what's required vs optional and the default value when optional. e.g., `anonymous` is optional, default `false`

src/helpers/email.py
```diff
@@ -0,0 +1,95 @@
...
+import os
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+
+SMTP_USER = os.environ.get("SMTP_USER", "")
+SMTP_PASSWORD = os.environ.get("SMTP_PASSWORD", "")
+SMTP_HOST = os.environ.get("SMTP_HOST", "smtp.gmail.com")
+SMTP_PORT = int(os.environ.get("SMTP_PORT", "587"))
```
houtanb, line 18: do that in `src/helpers/env.py`

houtanb (changes requested):

PR #213, opened 2026-06-09
refactor: fred

nikbpetrov:
- the hard-coded fred qs dict is in a separate file for maintainability


nikbpetrov:
parity tests pass, with both new jobs ever so slightly faster, and only minor deviations between old/new output vs prod nightly prod outputs (minor numeric differences, expected)

nikbpetrov:
Tests pass again

src/sources/_metadata.py
```diff
@@ -53,6 +55,9 @@
                 nullification_start_date=date(2025, 11, 1),
             ),
         ],
+        # Predefined FRED series the fetch job pulls each run (data lives in the
+        # lightweight _fred_questions module so consumers needn't import sources.fred).
+        "questions": FRED_QUESTIONS,
```
houtanb: @nikbpetrov is this used anywhere? Why add this to this file? I think `fred_questions` is only used by fetch, right? Totally fine to keep it as a separate file, but then why not just import it into `fred.py` as opposed to adding it to the metadata? Essentially treat it like `_METEOFRANCE_STATIONS` in `dbnomics.py` while keeping it in a separate file for maintainability.

houtanb (changes requested):

src/sources/_metadata.py
nikbpetrov: `self.questions` in `sources/fred.py` that avoids the import but you're right that import into fetch directly makes more sense

nikbpetrov:
tests pass

PR #216, opened 2026-06-10
refactor: dbnomics

nikbpetrov:
both fetch and update jobs do the sinful df concatenating instead of gathering result objects in a list and then running a single `pd.DataFrame()` over it, but have kept it as is, as the performance improvement is negligible, and at least now it preserves parity


nikbpetrov:
Tests pass w/ perfect parity.

nikbpetrov:
Tests pass again (old/new matches perfectly), though with predictable differences to prod: there was one more observation published between 00:08:50 when the prod snapshot happened vs when the tests ran today (10ish hours later)

PR #219, opened 2026-06-14
refactor: wikipedia

nikbpetrov:
there are plenty of decisions I made here along the way, so if anything is not clear, let me know


nikbpetrov:
Tests pass w/ perfect parity.

houtanb:
~~@nikbpetrov we plan on dropping Wikipedia, so not worth focusing more on this atm. I'll put off reviewing but leave the PR open in case I find some free time for this~~

Edit: retracting that. It'll still be good to refactor this as we'll be pulling the data in for the foreseeable future

PR #220, opened 2026-06-15
Implementing a dynamic resolution task counter

### LINKED ISSUE #202: Remove the hard-coded number of parallel tasks for resolution
The number of parallel tasks for resolution is hard coded to $50$, with the extra tasks being wound down.

The relevant code line is in [src/nightly_update_workflow/worker/main.py#17](https://github.com/forecastingresearch/forecastbench/blob/830746413ab1ba2ba51aa506e887ba99619ba7a9/src/nightly_update_workflow/worker/main.py#L17):
```python
resolve_forecasts = [
    [
        ("func-resolve-forecasts", True, cloud_run.timeout_1h * 3, 50),
    ]
]
```

This was a short-term hack.

Fix this by:
* counting the number of folders in the forecast sets bucket
* launching that many tasks instead of $50$

oskariluoma:
closes #202

### DIFF
--- a/src/nightly_update_workflow/worker/main.py
+++ b/src/nightly_update_workflow/worker/main.py
@@ -4,7 +4,8 @@ import os
 import sys
 
 sys.path.append(os.path.join(os.path.dirname(__file__), "../.."))
-from helpers import cloud_run, dates, question_curation  # noqa: E402
+from helpers import cloud_run, dates, env, question_curation  # noqa: E402
+from orchestration import _io  # noqa: E402
 from sources import ALL_SOURCE_NAMES, SOURCE_METADATA  # noqa: E402
 
 metadata = [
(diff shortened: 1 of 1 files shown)


src/nightly_update_workflow/worker/main.py
```diff
@@ -4,19 +4,27 @@
...
-    ]
-]
+
+
+def get_resolve_forecasts():
+    """Build the resolve_forecasts job list with one task per forecast set folder."""
+    _, date_folders = _io.get_valid_forecast_files_and_dates(bucket=env.FORECAST_SETS_BUCKET)
+    task_count = len(date_folders)
+    if task_count == 0:
+        print("WARNING: No forecast set folders found in bucket, running 1 task.")
```
houtanb: this won't happen but, if it did, it should raise an error

houtanb (changes requested):

houtanb:
@Oskaril-data also note the CI fails (run `make lint`)

PR #223, opened 2026-06-18
fix: delete `probability` in yfinance fetch

nikbpetrov:
## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the obsolete `probability` field from Yfinance fetch/update outputs to align with the current results contract.
  * Updated delisted-ticker flow so resolved records keep the expected question data and `freeze_datetime_value` behavior after the field removal.
* **Tests**
  * Updated test fixtures and assertions to stop expecting `probability`, while continuing to validate `freeze_datetime_value` and resolved-row contents.

### DIFF
--- a/src/_schemas.py
+++ b/src/_schemas.py
@@ -117,10 +117,9 @@ class PolymarketFetchFrame(QuestionFrame):
 
 
 class YfinanceFetchFrame(QuestionFrame):
-    """Output of YfinanceSource.fetch(). QuestionFrame plus transient fields for update()."""
+    """Output of YfinanceSource.fetch(). QuestionFrame plus a transient field for update()."""
 
     fetch_datetime: Series[str]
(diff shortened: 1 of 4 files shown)


nikbpetrov:
@houtanb note that this change means that there will be a stale `probability` fields across all files, should I provide a command for you to run against prod to remove that field across all files? very destructive action, and I will test ofc (and provide testing guidance for the dev project)

houtanb:
@nikbpetrov where else is the probability field perpetuated beyond the fetch file? Won't it be removed after one run?

nikbpetrov:
yup, sorry, that's right -- my testing harness threw me off as it also surfaced errors on `update()` -- but that was a harness problem, not a PR problem

houtanb:
@coderabbit full review

houtanb:
merged in d2b3364

PR #227, opened 2026-06-18
fix: force renamed tickers files' to be identical by constructions; avoid fetching the old ticker name unnecessarily

nikbpetrov:
on top of #225


houtanb:
merged in 30b5122e69c3eb6d874717c765035cc8b950e3f6

PR #230, opened 2026-06-20
feat: harmonize deployment to use dockerfiles and bump to `google-24-full` and `python314`

### LINKED ISSUE #222: refactor: makefile cleanup
Some Makefiles use `Procfile` and some use `Dockerfile` to deploy.

The Dockerfiles are using Python 3.12 but most (if not all) of the code can run with Python 3.14. Update to Python 3.14, potentially simplify to use ~~Procfile~~ Dockerfiles

nikbpetrov:
Resolves #222 

Tested together with #231


houtanb:
@nikbpetrov the above is a big improvement. Could you make these changes in this commit as well:
1. Add `deploy_config.mk` (or other name) and consolidating the `.gcloudignore` and `Dockerfile` rules there and simply import that everywhere to consolidate the cookie-cutter build rules
1. Add `.python-version` to the root dir and:
    1. derive `PYTHON_RUNTIME` from that in the Makefile 
    1. use it to set the python version for thein the CI: `.github/workflows/makefile.yml`. Claude says you get this for free if you move to `actions/setup-python@v7`

nikbpetrov:
Done - very clean indeed.

2ii verified [here](https://github.com/actions/setup-python#basic-usage) with `python-version-file` arg optional.

set `.DEFAULT_GOAL := lint` in `Makefile` (root one) just becasuse we currently have `include *.mk` which defines a bunch of vars in there - it's just comestic noise but without that default goal the first included command would be executed on an empty `make` call

.github/workflows/makefile.yml
houtanb: can this also go to `@v7`?

houtanb:
Great! One last thing to test out: edit `deploy_config.mk` to be like:
```makefile
DOCKERFILE_TEMPLATE ?= $(ROOT_DIR)src/helpers/Dockerfile.template
Dockerfile: $(DOCKERFILE_TEMPLATE)
```

then in move `src/nightly_update_workflow/compress_buckets/Dockerfile` to `src/nightly_update_workflow/compress_buckets/Dockerfile.template` and then in the `Makefile` in that dir do:
```makefile
DOCKERFILE_TEMPLATE = Dockerfile.template
include $(ROOT_DIR)deploy_config.mk
```
To be tested, but I think that would let us get rid of the hard-coded serverless runtiime and python version.

If you agree the above is better, could you make the fix and also do a complete test run to make sure it all goes through? There might be a problem with Kalshi at the moment as it broke last night, so might need to skip that.

PR #236, opened 2026-06-23
refactor: rewrite LLM forecaster

houtanb:
Align with recent updates to the `utils` repo. Remove the `utils` submodule accordingly.

Keep a log, both in markdown and jsonl of all prompts and responses for a model run.

Handle changes in leaderboard as well, mapping legacy models to the new model naming structure and better handling model release dates.

Change the reformat model to gpt-5-mini.


nikbpetrov:
Just flagging that the PR is defo doing more than one thing (unrelated) - still going through, but I can see the orchestration upload change, the utils package deploy change and within llm forecater-related changes, i think there are two parts, but i am still not sure how intertwined they are

houtanb:
Agreed, it's a monster.

* the llm forecaster is what was preventing `utils` from being incorporated used as a package in the repo and instead of coding it as a submodule and making a second commit pulling it out I combined them because it made coding _much_ simpler and avoided a follow-on commit that undid a lot of what the initial commit did.
* the build changes are related because of `requirements.runtime.txt`. As part of adding that, I simplified the way build rules worked to avoid inconsistencies across Makefiles.

Feel free to focus primarily on the llm-forecsater-related changes (`src/llm_forecaster`, `src/orchestration/func_llm_forecaster_*`

AGENTS.md
```diff
@@ -102,6 +109,17 @@ NB: `variables.mk` contains both runtime environment variables and variables tha
 * Run `make lint` to apply formatting and check all of the above
 * Configuration is in `pyproject.toml` and `setup.cfg`
 * Functions should use type hints
+* New ForecastBench Python files should not add `from __future__ import annotations`
+* Prefer f-strings for Python string interpolation, including log messages. Use logger
+  %-style only when there is a concrete, necessary reason to defer interpolation.
+* Tests should verify behavior, data contracts, or externally meaningful integration points. Do
+  not add tests that only assert implementation details such as exact helper signatures, private
+  call shapes, or the absence/presence of internal arguments unless that detail is an explicitly
+  supported public API.
```
nikbpetrov, line 118: I am not so sure about this one. I would like to have unit tests that test a specific function's signature, private OR public. It's precisely the private funcs that are likely to let edge case errors seep through.

houtanb: your coder may be better than mine. I was getting a bunch of junk tests that were not very interesting and this helped. But I'll go on the record and say most of the tests here are junk and not thought out at all, hence the necessity of spending time on actually developing tests as opposed to using the LLM test output

src/llm_forecaster/prompts.py
```diff
@@ -0,0 +1,143 @@
...
+Resolution Criteria:
+{resolution_criteria}
+
+Today's Date: {today_date}
+
+Question resolution date: {resolution_date}
+
+Output your answer (a number between 0 and 1) with an asterisk at the beginning and end of the decimal.
+Do not output anything else.
+Answer: {{ Insert answer here }}
```
nikbpetrov, line 62: is there any reason why we are not using [structured outputs](https://developers.openai.com/api/docs/guides/structured-outputs) where an LLM provider supports it? most do nowadays

nikbpetrov: +1 actually, after looking through `runner.py` and `parsing.py` - that latter file becomes largely superfluous with structured outputs

houtanb: That and potentially batch processing are coming in a second step because they involve a bit more work, finding which providers support it and supporting only for those, testing

src/llm_forecaster/model_run_transcripts.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import Any
+
+from llm_forecaster import fb_model_runs, output
+from llm_forecaster.forecast_variants import ForecastVariant
+from orchestration import _io
+
+logger = logging.getLogger(__name__)
+
+
+class LLMCallTranscript:
```
nikbpetrov, line 21: This is nice, but I don't understand why it would be there? Would this be human-readable? I guess that's okay...

But more importantly, why would we **not** want to keep a logfile of the raw HTTP requests? If I were a very picky reviewer/audior, I would want to go through the raw requests, see the model call's `fingerprint`, reason(s) for rejecting to answer (eg out of tokens) and more. Crucially, especially with agents, these humongous logfiles are readily parsable and analyzable. At worst, we can add a pruning job to avoid overrun, though with increased importance of FB, these historic snapshots might prove ever more valuable.

nikbpetrov:
I guess this would not be a comment for this PR, but does relate to it and utils...

In utils, gettign responses from models uses each provider's SDK (which, to be fair, I am a tad surprised given the general tendency to rely on fewer packages/dependencies, especially some of those SDKs being a tad questionably maintained). However, this prevents some possible cost optimizations. OpenAI offers 50% reduced bill with the Batch API, Deepseek has times of day where requests cost less etc. While this is some infra for us to stand up, the cost reductions might very well be worth it.

src/llm_forecaster/fb_model_runs.py
```diff
@@ -0,0 +1,102 @@
...
+    "grok-4.20-0309-reasoning-run-variant-01",
+    "grok-4.20-0309-non-reasoning-run-variant-01",
+    "grok-4.3-run-variant-01",
+    "gemini-3.5-flash-run-variant-01",
+    "gemini-3.5-flash-run-variant-02",
+    "gemini-3.1-pro-preview-run-variant-01",
+    "gemini-3.1-flash-lite-run-variant-01",
+]
+
+FB_MODEL_RUNS = shared_model_runs.select_model_runs(FB_MODEL_RUN_KEYS)
```
nikbpetrov, line 47: shouldn't this be failing? 5 model runs are missing from [utils/model_runs](https://github.com/forecastingresearch/utils/pull/26/changes#diff-92f843bcb21ba3ae26f9213f884a4d663d38f1171535f51ea6c48b5c4714d27a) and in that file lines 1140-1142 index the dict directly, instead of with a defensive `.get()`

minimax-m2.7-run-variant-02, minimax-m3-run-variant-01, glm-5.2-run-variant-01, claude-opus-4-8-run-variant-05, gemini-3.5-flash-run-variant-02

src/llm_forecaster/model_run_transcripts.py
houtanb: Doing this for two reasons:
* markdown: to easily ensure the prompt is expanded to what was intended
* jsonl: to track prompt and model responses to ensure model extraction is done correctly (easier to work with programmatically later). 

But this it's a great idea to add logging at the provider level for raw calls / responses for future audits!

houtanb:
> I guess this would not be a comment for this PR, but does relate to it and utils...
> 
> In utils, gettign responses from models uses each provider's SDK (which, to be fair, I am a tad surprised given the general tendency to rely on fewer packages/dependencies, especially some of those SDKs being a tad questionably maintained). However, this prevents some possible cost optimizations. OpenAI offers 50% reduced bill with the Batch API, Deepseek has times of day where requests cost less etc. While this is some infra for us to stand up, the cost reductions might very well be worth it.

See the comment [here](https://github.com/forecastingresearch/forecastbench/pull/236#discussion_r3466047237)

AGENTS.md
nikbpetrov: i doubt it re code: both opus and codex feel lobotomized in recent week(s)

src/llm_forecaster/parsing.py
```diff
@@ -0,0 +1,124 @@
...
+        if isinstance(value, bool) or not isinstance(value, int | float):
+            raise ValueError(f"{_EXTRACTION_ERROR}, got: {raw_response!r}")
+        if not 0 <= value <= 1:
+            raise ValueError(f"{_EXTRACTION_ERROR}, got: {raw_response!r}")
+        probabilities.append(float(value))
+
+    return probabilities
+
+
+def _extract_forecast_probabilities(
```
nikbpetrov, line 64: the fallback to LLM extraction is definitely an interesting choice; I can see the point in it, but I would NOT trust it - not even close!

I'd want a clear one-stop-shop to see every single fallback extraction (raw input -> forecast) at the very least, if not even slack msgs (at least in the early days) to quickly check (perhaps toggle-able with an ENV var)

nikbpetrov: bonus points if we can figure out why the static extraction might have failed; i've found in the past that usually if it fails, it's because a given LLM has a specific 'style'

e.g. once i was asking LLMs to assume a specific persona description and was extracting a number and the extraction worked in all models except some flavour of the OpenAI ones, where it decided to always start its responses with "As XX-year-old xxxx", which was throwing it off.

houtanb: This is what the markdown/.jsonl file is for and how I have tested the extraction.

Extraction by regular expression works for dumb models. Reasoning models tend to output extra info, e.g.,:
```text
I'll research the current situation regarding Guam and any relevant geopolitical tensions before making my predictions.Let me fix the parsing of search results.Based on my research, I can now make an informed assessment.

## Analysis

**Key facts:**
- The reference threshold is **0.0** — Guam has had **zero** ACLED 'Explosions/Remote violence' events over the preceding 360 days. This means *any single event* (≥1) in a 30-day window resolves the question TRUE.
- Guam is a heavily defended U.S. territory with essentially no history of recorded explosions/remote violence events.

**Current situation (mid-2026):**
- North Korean missile launches continue, but Guam authorities consistently assess <cite index="0-0">"No immediate threat was assessed for Guam or the Marianas from the reported launch."</cite> Tests splash down elsewhere and are coded by ACLED at those locations, not Guam.
- China-Taiwan conflict is a modeled risk, with think tank scenarios <cite index="6-3">assumed to occur between 2026 and 2030</cite>, but no active conflict exists.
- The U.S. is heavily investing in defenses — <cite index="13-0">Lockheed Martin secured contracts raising total value to $1.9B to defend U.S. Pacific bases against China missile threats.</cite>
- The notable 2026 "attack" on Guam was a **cyber** incident, which does not count as ACLED 'Explosions/Remote violence'.

**Reasoning:**
- A genuine explosion/remote violence event in Guam would essentially require an act of war (North Korean strike, or Chinese attack during a Taiwan conflict). These are very low-probability for any specific 30-day window.
- Near-term probabilities are minimal (status quo, no active conflict). Longer horizons carry slightly elevated risk given the 2027–2030 Taiwan risk window and North Korea's developing strike capabilities, but a single 30-day window catching active hostilities remains unlikely.

## Predictions

For 2026-06-28: *0.005*
For 2026-07-21: *0.006*
For 2026-09-19: *0.008*
For 2026-12-18: *0.012*
For 2027-06-21: *0.018*
For 2029-06-20: *0.028*
For 2031-06-20: *0.033*
For 2036-06-18: *0.04*
```

I agree that structured output, where it's supported, is the way to go. This prompt seems to be working well for extraction. Another idea I had is to test the extraction prompt against two models and abort if they don't both return the same thing.

nikbpetrov:
> > I guess this would not be a comment for this PR, but does relate to it and utils...
> > In utils, gettign responses from models uses each provider's SDK (which, to be fair, I am a tad surprised given the general tendency to rely on fewer packages/dependencies, especially some of those SDKs being a tad questionably maintained). However, this prevents some possible cost optimizations. OpenAI offers 50% reduced bill with the Batch API, Deepseek has times of day where requests cost less etc. While this is some infra for us to stand up, the cost reductions might very well be worth it.
> 
> See the comment [here](https://github.com/forecastingresearch/forecastbench/pull/236#discussion_r3466047237)

Fair, though let me double-down on the SDKs: why? why not raw HTTP requests? In our case they are pretty simple. That way we could wrap those requests within the one-day-coming HTTP wrapper?

It'd be a tedious rewrite right now as this is now tested and works, but still curious more generally, whether it was a toin coss or there was a more specific reason.

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,616 @@
...
+    ZERO_SHOT,
+    ZERO_SHOT_WITH_FREEZE_VALUES,
+    ForecastVariant,
+)
+from llm_forecaster.model_run_transcripts import (
+    LLMCallTranscript,
+    TranscriptRecordingModelRun,
+)
+from llm_forecaster.question_set import QuestionSet, split_questions
+from orchestration import _io
```
nikbpetrov: ops, was almost going to miss this one! quite the illegal move I'd argue

As per our refactor plan, `src/` cannot import from `orchestration/` - io lives at the boundary; refactored sources already enforce this by outputting files and then the boundary takes care of uploading them (which is even faster)

nikbpetrov: applies to `question_set.py` and `model_run_transcripts.py`

nikbpetrov: i guess buffering these files can be expensive (in case of an unhandled error, everything burns within a job) but useful to handle: forces us to figure out how to handle problems robustly OR forces a generator pattern for responses yielding

pyproject.toml
```diff
@@ -1,13 +1,16 @@
 [tool.pytest.ini_options]
 testpaths = ["src/tests"]
 pythonpath = ["src"]
+filterwarnings = [
```
nikbpetrov: jobs are all on 3.12?

src/tests/llm_forecaster/smoke_test/smoke_test.py
```diff
@@ -0,0 +1,398 @@
...
+    """Return a compact row-count mismatch message for partial smoke outputs."""
+    row_counts = [
+        f"{written_file.local_filename} has {len(written_file.rows)}"
+        for written_file in written_files
+        if len(written_file.rows) != expected_rows
+    ]
+    return f"Expected {expected_rows} rows per forecast file; {'; '.join(row_counts)}."
+
+
+def run_smoke_test(
```
nikbpetrov, line 246: should this file be under `src/` - it's a good one but it's no `src/`-worthy; something like `tests/`, though with GCP-dependence, I am not sure if it's not relegatable to a different repo - up to you, but defo not `src/`

houtanb:
> > > I guess this would not be a comment for this PR, but does relate to it and utils...
> > > In utils, gettign responses from models uses each provider's SDK (which, to be fair, I am a tad surprised given the general tendency to rely on fewer packages/dependencies, especially some of those SDKs being a tad questionably maintained). However, this prevents some possible cost optimizations. OpenAI offers 50% reduced bill with the Batch API, Deepseek has times of day where requests cost less etc. While this is some infra for us to stand up, the cost reductions might very well be worth it.
> > 
> > 
> > See the comment [here](https://github.com/forecastingresearch/forecastbench/pull/236#discussion_r3466047237)
> 
> Fair, though let me double-down on the SDKs: why? why not raw HTTP requests? In our case they are pretty simple. That way we could wrap those requests within the one-day-coming HTTP wrapper?
> 
> It'd be a tedious rewrite right now as this is now tested and works, but still curious more generally, whether it was a toin coss or there was a more specific reason.

The `utils` provider adapters already used SDKs so I kept this PR aligned with the existing implementation. We can always change this at a later date if it becomes a priority as the boundary is well-defined on each provider.

Batch/off-peak optimization feels like a related but separate execution-mode follow-up; we'd need to do a lot more than swap out the SDK to make batch processing work.

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,616 @@
...
+
+def _forecast_dataset_questions(
+    model_run: fb_model_runs.ModelRun,
+    dataset_questions: list[dict[str, Any]],
+    forecast_due_date: str,
+    today_date: str,
+    variant: ForecastVariant,
+    transcript: LLMCallTranscript,
+    raise_on_question_error: bool = False,
+) -> list[dict[str, Any]]:
```
nikbpetrov: We have in the plan a type for this:

```python
class ForecastFrame(pa.SchemaModel):
    """What forecasters submit (and what naive/dummy forecasters produce)."""
    id: Series[object]          # str or tuple for combo
    source: Series[str]
    forecast: Series[float]
    resolution_date: Series[str]
    reasoning: Series[str]

    class Config:
        strict = False
        coerce = False
```

there are a couple of places where returning this could be useful; then, at whatever ending boundary we have, we should probably also do a `validate()`

src/llm_forecaster/parsing.py
nikbpetrov: yup, calibrating is one way; I did something like this for another project where we tested 3 different models on performing an extraction-like task, and picked the 3-model combo that achieved higher accuracy (only using the third one if a tie-breaker is needed)

but that becomes plenty of extra infra to stand up

again, the extraction model is fine, just easier auditing would make it a tad safer

nikbpetrov:
Apart from the issues listed above, I can't see anything else for the time being. It's a really massive PR and while I could sink more time into it, it looks like it's in a solid shape overall. Reuses many existing patterns in terms of orchestration (with some fine-tuning of those patterns, and that fine-tuning appears solid, and easily-testable so a few dev bucket runs should catch issues easily), so it's mainly the internal wiring of the model runs that's potentially dangerous. In general, logging and error handling is a bit messy but I've chased through a few threads and they do seem to hold up in the end.

If you can, I'd run this entire dev pipeline in parallel with the prod one for the next LLM forecasting round and compare, just to be on the safe side.

src/llm_forecaster/fb_model_runs.py
```diff
@@ -0,0 +1,102 @@
...
+FB_MODEL_RUNS_BY_KEY = {run.model_run_key: run for run in FB_MODEL_RUNS}
+FB_MODEL_RUNS_BY_SLUG = {run.slug: run for run in FB_MODEL_RUNS}
+FORECAST_EXTRACTION_MODEL = shared_model_runs.get_model_run("gpt-5-mini-2025-08-07-run-variant-01")
+
+
+PROVIDER_MAX_WORKERS = {
```
elsehow, line 54: This will `KeyError` at runtime in the worker if any selected run's provider isn't one of the five keys. That's fine today, but feels footgun-y (if someone adds a run on a provider not in this dict later). Consider `.get(provider, <default>)`.

src/llm_forecaster/output.py
```diff
@@ -0,0 +1,185 @@
...
+
+def forecast_file_data(
+    forecast_due_date: str,
+    question_set_filename: str,
+    model_run: ModelRun,
+    variant: ForecastVariant,
+    rows: Sequence[Mapping[str, Any]],
+) -> dict[str, Any]:
+    """Return a ForecastBench LLM forecast file using the current identity schema."""
+    uses_freeze_values = variant.uses_freeze_values
```
elsehow: I/Claude may be misunderstanding this, but it looks like (a) the dataset prompt always includes the freeze value and (b) both output variants share one dataset pass. That means the `uses_freeze_values: false` file still has dataset forecasts made *with* the freeze value. That is, the flag really only describes market methodology. Is this intended behavior?

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,616 @@
...
+    dataset_expected_count: int,
+) -> None:
+    """Log final successful forecast counts by market variant and dataset rows."""
+    for variant_group in DATASET_FORECAST_SHARING_VARIANT_GROUPS:
+        for variant in variant_group.output_variants:
+            summary_label = _forecast_summary_label(label=variant.key)
+            logger.info(
+                f"LLM forecast success {summary_label}: "
+                f"{market_success_counts[variant]}/{market_expected_count}"
+            )
```
elsehow, line 416: If we ever add a second `DatasetForecastingSharingVariantGroup`, the dataset would get re-forecast per group, which means extra LLM spend the "sharing" name implies we want to avoid. Also, the `dataset_success_count in `run_model` will be overwritten each group, so `_log_forecast_success_counts` would only report the last group's dataset count.

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,616 @@
...
+        model_run=model_run,
+        is_test=is_test,
+    )
+
+
+def _formatted_question(question: dict[str, Any], forecast_due_date: str) -> str:
+    return question["question"].format(
+        forecast_due_date=forecast_due_date,
+        resolution_date="each of the resolution dates provided below",
+    )
```
elsehow, line 75: This method injects `resolution_date="each of the resolution dates provided below"` into market question text too, right? For a single-horizon market question, that plural phrasing is odd if the question template embeds `{resolution_date}`. Can we confirm that the phrasing is acceptable there?

elsehow (approved):
This looks great! I have a few very minor notes, but nothing I consider blocking. Thank you for doing this!

src/llm_forecaster/fb_model_runs.py
houtanb: those models are present (your link is to an old commit). see this [link](https://github.com/forecastingresearch/utils/pull/26/changes#diff-d4f5111e7422e2d33c6dc96a1d0b0143542b5ef2).

also `make test` fails if a model run key is not present. Though I took this opportunity to simplify the test for this, thank you!

src/llm_forecaster/runner.py
houtanb: Thanks for this. I went for the generator pattern!

pyproject.toml
houtanb: Good point. The CI/deployed jobs are on Python 3.12, so this filter was only quieting my local Python 3.14 environment. It doesn’t belong in shared `pytest` config for this PR; I’ll remove it.

src/tests/llm_forecaster/smoke_test/smoke_test.py
houtanb: We can move out of `src/llm_forecaster`. I’d avoid adding root `tests/` since this repo already has `src/tests/`. Because this is a manually run smoke/integration command rather than a normal `pytest` test, I’m leaning toward `scripts/llm-forecaster-smoke-test/smoke_test.py`. Open to `src/tests/llm_forecaster/smoke_test/smoke_test.py` if you’d rather group it with test assets, but `scripts/` seems cleaner to me. Thoughts @nikbpetrov

nikbpetrov: do `src/tests/`, that's fine, I am planning something similar for the testsuite i am working on; it will probably go into something like `src/tests/harness/` along other harness-specific/admin tests

src/llm_forecaster/runner.py
nikbpetrov: good choice that I see this again: you could potentially parallelize this across all llm runs in a monolith too (instead of a GCP job per llm run) - arguably faster and more reliable for handling timeouts/requests limits; might be quite nice with bigquery or sth else that would allow us to ensure partial data saving reliably

houtanb:
> Apart from the issues listed above, I can't see anything else for the time being. It's a really massive PR and while I could sink more time into it, it looks like it's in a solid shape overall. Reuses many existing patterns in terms of orchestration (with some fine-tuning of those patterns, and that fine-tuning appears solid, and easily-testable so a few dev bucket runs should catch issues easily), so it's mainly the internal wiring of the model runs that's potentially dangerous. In general, logging and error handling is a bit messy but I've chased through a few threads and they do seem to hold up in the end.
> 
> If you can, I'd run this entire dev pipeline in parallel with the prod one for the next LLM forecasting round and compare, just to be on the safe side.

That's the plan! :)

src/llm_forecaster/runner.py
houtanb: So many ways to do this! :)

src/llm_forecaster/output.py
houtanb: yes that's intended. The dataset prompt is always the same, only the market prompt changes to have zero shot and zero shot with freeze values. I'll rename the variable to `market_prompt_uses_freeze_values` to make it more clear.

src/llm_forecaster/runner.py
houtanb: Oh, yeah, but that would be a different functioning than intended. This setup supports what we've done in the past, e.g. for scratchpad prompts. There was a general prompt variant, and the market prompts differed in that they provided or did not provide the crowd forecast. But the forecasts generated dataset scratchpad prompt would only be shared with the scratchpad and scratchpad with freeze value outputs, not with another variant group.

src/llm_forecaster/runner.py
houtanb: yeah, this is a bit misleading. It only actually applies to dataset questions, not market questions, as those pull the "question" field from the prediction platform and lack these placeholders. Will make that more explicit, thanks!

src/llm_forecaster/runner.py
```diff
@@ -591,26 +619,40 @@ def run_model(
...
                 rows=rows,
-                upload=upload,
                 is_test=is_test,
             )
-            written_files.append(written_file)
             logger.info(
                 f"Wrote LLM forecast file variant={variant.key} "
                 f"rows={len(written_file.rows)} forecast_file={written_file.local_filename}"
             )
+            yield written_file
```
nikbpetrov: Nice! Though, I still don't understand the logic why the IO is still here in `src/llm_forecaster`. The file itself contains rows (in a specific shape) - that's it. If we switch to a db setup, we'd have to rewrite all this io logic. Why not have the iterator return the rows/data and then how these are saved all oges in `_io.py` (or a new file, something in `orchestration`). Notice the sources too with the resolution files: the 'resolution_files' is actually just a `dict[str, pd.DataFrame]` which gets passed onto the orchestration/driver, which handles those however it wants. It's fine to have a name method or similar (something that can be deleted wholesale if io pattern changes), similar to acled/wikipedia's `hash_mapping`s , but not wholesale writing. If we changed this to db tomorrow, `destination_blob_name` (or any `Path`s) would have to get torn away and that's going to be a tad more painful than necessary I think.

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,693 @@
...
+
+    return _forecast_questions(
+        model_run=model_run,
+        questions_to_forecast=market_questions,
+        forecast_question=forecast_question,
+        progress_label=variant.key,
+    )
+
+
+def _validated_forecast_frame_from_rows(
```
nikbpetrov, line 371: a lot of validating happening here - validation is in general rather expensive and should happen only at the public boundaries

oh, actually, while `_validated_forecast_frame_from_rows` seems acceptable as it happens once after all the threads have run (so that'd be just the once per job/llm run I think), `_sorted_forecast_rows` is happening mid loop - why?

src/llm_forecaster/runner.py
```diff
@@ -0,0 +1,693 @@
...
+    rows: Sequence[dict[str, Any]],
+) -> DataFrame[ForecastFrame]:
+    frame = pd.DataFrame(rows, columns=list(ForecastFrame.to_schema().columns))
+    if frame.empty:
+        frame["forecast"] = frame["forecast"].astype(float)
```
nikbpetrov: don't allow returning empty frames like that and those hacky checks/casts disappear - return `None` when df empty, consistent with other sources/refactor work - those get thrown away anyway inside the loop and empty dfs are not actually written

PR #239, opened 2026-06-26
Lazyfying

nikbpetrov:
meant to be merged as separate commits, just grouped here in the same PR

7de71bd is not lazy-related, sorry for sneaking it in, but it's a 1-line print -> logger.debug, change (preps the testing framework work)

### DIFF
--- a/src/curate_questions/create_question_set/main.py
+++ b/src/curate_questions/create_question_set/main.py
@@ -15,7 +15,6 @@ Market question sampling aims to achieve balanced representation across:
 import json
 import logging
 import os
-import random
 import sys
 from collections.abc import Callable
 from copy import deepcopy
@@ -23,6 +22,7 @@ from datetime import datetime, timedelta
 from enum import Enum
 from fractions import Fraction
 
+import numpy as np
 import pandas as pd
 from tqdm import tqdm
 from utils import gcp
@@ -94,15 +94,19 @@ def process_questions(
     single_generation_func: Callable,
     show_plots: bool,
     question_set_target: QuestionSetTarget,
+    random_state=None,
 ) -> dict:
     """Sample from `questions` to get the number of questions needed.
 
     Args:
         questions (dict): Source questions keyed by source name, each with a "dfq" DataFrame
         to_questions (dict): Allocation info keyed by source with "num_questions_to_sample"
-        single_generation_func (Callable): Sampling function taking (values, n) and returning DataFrame
+        single_generation_func (Callable): Sampling function taking (values, n, random_state) and
+            returning a DataFrame
         show_plots (bool): Whether to display distribution plots
         question_set_target (QuestionSetTarget): Target question set ("llm" or "human")
+        random_state: Seed/``np.random.RandomState`` threaded to ``single_generation_func`` for
+            reproducibility. ``None`` (default) samples without a fixed seed.
 
     Returns
         processed_questions (dict): Deep copy of questions with sampled DataFrames
@@ -118,7 +122,7 @@ def process_questions(
         df_available = values["dfq"].copy()
 
         # Sample questions for this source
-        values["dfq"] = single_generation_func(values, num_single)
+        values["dfq"] = single_generation_func(values, num_single, random_state)
         df_sampled = values["dfq"]
         num_found += len(df_sampled)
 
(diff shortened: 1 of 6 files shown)


src/helpers/keys.py
```diff
@@ -1,9 +1,38 @@
...
 
+# Public attribute name -> Secret Manager secret name. The two differ only for GOOGLE/GEMINI.
+_SECRET_NAMES = {
+    # LLM
+    "API_KEY_ANTHROPIC": "API_KEY_ANTHROPIC",
+    "API_KEY_OPENAI": "API_KEY_OPENAI",
+    "API_KEY_TOGETHERAI": "API_KEY_TOGETHERAI",
+    "API_KEY_GOOGLE": "API_KEY_GEMINI",
+    "API_KEY_MISTRAL": "API_KEY_MISTRAL",
+    "API_KEY_XAI": "API_KEY_XAI",
```
houtanb, line 19: these were removed in the llm forecaster rewrite, added back here. Why?

src/leaderboard/main.py
```diff
@@ -2639,30 +2668,16 @@ def generate_simulated_leaderboards(
...
-        return retval.drop(columns=["draw"])
-
     def bootstrap_and_score(idx):
         logger.info(f"[replicate {idx+1}/{N}] starting...")
+        # Per-replicate RandomState so each loky child is deterministic when `seed` is set.
+        random_state = None if seed is None else np.random.RandomState(seed + idx)
         df_bs = (
             df.groupby(["forecast_due_date", "source"])
-            .apply(question_level_bootstrap, include_groups=False)
+            .apply(
```
houtanb, line 2677: nit: from the [docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.typing.DataFrameGroupBy.apply.html) so can you just drop the `lambda` and do `.apply(question_level_bootstrap, include_groups=False, random_state=random_state)`

src/curate_questions/create_question_set/main.py
```diff
@@ -1252,6 +1281,12 @@ def driver(_: None) -> None:
         )
         HUMAN_QUESTIONS.update(human_questions_of_question_type)
 
+    # Optional reproducibility: when QUESTION_SET_SEED is set, thread a single RandomState through
```
houtanb, line 1284: Can you explicitly mention "for testing"? Claude was confused about this and thought the seed was missing in deployment.

src/curate_questions/create_question_set/main.py
```diff
@@ -1252,6 +1281,12 @@ def driver(_: None) -> None:
...
+    # both sampling passes so the published set is deterministic. Unset (the default) preserves the
+    # historical unseeded behaviour.
+    seed_env = os.getenv("QUESTION_SET_SEED")
```
houtanb, line 1287: should we also check that `seed_env` is `None` when we're in prod to ensure it isn't accidentally ever set in deployment?

src/helpers/env.py
```diff
@@ -1,25 +1,47 @@
...
-LLM_BASELINE_SERVICE_ACCOUNT = os.environ.get("LLM_BASELINE_SERVICE_ACCOUNT")
-LLM_BASELINE_STAGING_BUCKET = os.environ.get("LLM_BASELINE_STAGING_BUCKET")
-LLM_BASELINE_NEWS_BUCKET = os.environ.get("LLM_BASELINE_NEWS_BUCKET")
-NUM_CPUS = int(os.environ.get("NUM_CPUS", 1))
-RUNNING_LOCALLY = bool(int(os.environ.get("RUNNING_LOCALLY", False)))
-BUCKET_MOUNT_POINT = os.environ.get("BUCKET_MOUNT_POINT", "")
-WORKSPACE_BUCKET = os.environ.get("WORKSPACE_BUCKET")
+# Names read as plain strings (``None`` if unset).
+_STR_VARS = {
+    "PROJECT_ID": "CLOUD_PROJECT",
```
houtanb, line 11: what about a commit renaming `PROJECT_ID` to `CLOUD_PROJECT` and simplifying this to a set of strings?

src/curate_questions/create_question_set/main.py
houtanb, line 1284: Instead of `QUESTION_SET_SEED` and the seed in the leaderboard code, what do you think about adding `RANDOM_SEED` to `env.py`? It would be an `int` and `None` by default and be referred to for any non-deterministic behavior in the codebase. Then the check for seed + prod can also be done in `env.py`

src/leaderboard/main.py
nikbpetrov: cool, new stuff, don't recall these back in the day when i was learning pandas

src/helpers/keys.py
nikbpetrov: missed it; when i rebase on top of main this is a huge insert, not a replace

src/curate_questions/create_question_set/main.py
nikbpetrov: not bad; but the only prod guard that makes sense to me is a test along the other runtime reqs; we dont have a really good dev setup here I dont think, so that's the only way so far (ideally we could check if the GCP project to which this is getting deployed is the actual prod one and if so, prevent `RANDOM_SEED` from getting deployed, as technically we could want to run tihs in a dev GCP project, but at the moment this seems more like a distinction w/o a difference)

src/curate_questions/create_question_set/main.py
nikbpetrov: ditto as above

src/helpers/keys.py
```diff
@@ -1,9 +1,31 @@
...
+    "API_KEY_FRED": "API_KEY_FRED",
+    # QUESTION MARKET SOURCES
+    "API_KEY_METACULUS": "API_KEY_METACULUS",
+    "API_KEY_POLYMARKET": "API_KEY_POLYMARKET",
+    # WORKFLOW BOT
+    "API_SLACK_BOT_NOTIFICATION": "API_SLACK_BOT_NOTIFICATION",
+    "API_SLACK_BOT_CHANNEL": "API_SLACK_BOT_CHANNEL",
+    # GITHUB
+    "API_GITHUB_DATASET_REPO_URL": "API_GITHUB_DATASET_REPO_URL",
+}
```
houtanb: Can this be a set? Seems like key and value are always the same

houtanb (commented):
I picked out 5479728a2 0e095847a and c09ad91c5 and will likely merge them after the test run on dev.

Can you make the `_SECRET_NAMES` change and squash the other deterministic seed commits into one?

Issue #181, opened 2026-04-23
Add Parity dates and question difficulty estimates to forecastbench-datasets repo


houtanb:
Every night, the expected Super parity dates shown on the Explore page are updated by the codebase.

Also question difficulty is estimated.

Track changes in https://github.com/forecastingresearch/forecastbench-datasets

PR #248, opened 2026-06-30
Fix/148, 181 resolution and leaderboard sets to one commit

oskariluoma:
`src/orchestration/_io.py` — upload_resolution_set now only uploads to the bucket (no git push). Added push_all_resolution_sets() that gathers all resolution sets from the bucket and pushes them in one commit.
`src/orchestration/func_push_resolution_sets/` — new Cloud Run job (func-push-resolution-sets) that runs push_all_resolution_sets() as a single task.
`src/nightly_update_workflow/manager/main.py` — runs the push job after resolve finishes, before leaderboards.
`src/nightly_update_workflow/worker/main.py` — registers the new job.
`Makefile` — adds the push-resolution-sets target and wires it into resolve.
`src/helpers/git.py` — clone_and_push_files skips empty commits and returns whether it pushed (so no false "pushed" log when nothing changed).

## Summary by CodeRabbit

* **New Features**
  * Added a new workflow step to push resolution sets after forecasts are resolved, before downstream updates run.
  * Introduced a job that collects all available resolution sets and publishes them in a single commit, helping avoid push conflicts.
  * Added deployment support for the new job so it can run in the cloud like existing workflow tasks.

* **Bug Fixes**
  * Skips creating empty commits when there are no file changes, while still allowing remote mirrors to catch up.
  * Improves error reporting for push failures.


src/helpers/git.py
```diff
@@ -84,6 +85,13 @@ def clone_and_push_files(
         shutil.copy(source, full_destination_path, follow_symlinks=False)
         repo.index.add([destination])
 
+    # Skip empty commits: nothing to push if the staged files match HEAD.
```
nikbpetrov: what happens if a mirror fails but git succeeds? mirror would never get the update on same-day rerun?

src/orchestration/_io.py
```diff
@@ -259,14 +260,52 @@ def upload_resolution_set(df: pd.DataFrame, forecast_due_date: str, question_set
...
+    Run as one job rather than having each parallel resolve task push its own set, which
+    avoids the race condition of concurrent pushes to the repository.
+    """
+    from helpers import git, keys  # noqa: E402
+
+    if env.RUNNING_LOCALLY:
+        logger.info("Running locally; not pushing resolution sets to git.")
+        return
+
+    folder = "datasets/resolution_sets"
```
nikbpetrov, line 276: probably good to add a trailing slash to ensure that we capture only that bucket (and not `datasets/resolution_sets_testing/` or whatever, especially in dev/staging envs

src/orchestration/func_push_resolution_sets/requirements.txt
```diff
@@ -0,0 +1,13 @@
+google-cloud-storage
+google-cloud-secret-manager
+pandas>=2.2.2,<3.0
+tqdm
+gcsfs==2025.7.0
+GitPython
+scipy
```
nikbpetrov: are all these needed?

nikbpetrov:
Apart from these minor notes, LGTM

src/helpers/git.py
Oskaril-data: Yeah you're right about that. My plan for the fix: only make a commit when something actually changed, but always push to both origin and the mirror anyway. Git only sends what's missing, so if a remote is already up to date the push just does nothing, and if the mirror fell behind it catches up on the next run. Would that work?

src/orchestration/func_push_resolution_sets/requirements.txt
Oskaril-data: I think GitPython is needed but the three others are not?

src/helpers/git.py
nikbpetrov: you're actually right about git, but then why would you even download from git then? git will indeed only push changes so even if older files changed somehow, those changes would be pushed correctly

no idea about mirrors, probably @houtanb can better advise; mirrors is written as a generic one so not sure what that _could_ capture in theory, and also not sure if that's how HF works either

src/orchestration/func_push_resolution_sets/requirements.txt
nikbpetrov: you might need the `google-cloud-*` ones, `gcsfs` and maybe `pandas`? defo `GitPython`

src/helpers/git.py
Oskaril-data: The diff check and the push do different jobs: git push only compares commit pointers, not file contents.

Two scenarios:
A: **content unchanged**. We clone origin, so HEAD == origin's HEAD. Diff has no changes => index.commit() is skipped => no new commit. origin.push() sees the remote already has the same commit HEAD => transfers nothing. But if yesterday's mirror push failed, the mirror is behind => the mirror push sends it the missing commits and it catches up. That's the fix for the original concern.

B: **content changed**. Diff has changes => new commit with index.commit() => both pushes send it, as expected.

Why not just always commit and let push sort it out? GitPython's index.commit() creates a commit even with an identical tree and origin.push() pushes it as the commit is new. Without the diff check, the dataset repo would gain an empty commit every night.

src/helpers/git.py
```diff
@@ -85,26 +86,26 @@ def clone_and_push_files(
...
+        if has_changes:
+            repo.index.commit(commit_message, author=author, committer=committer)
         origin = repo.remote(name="origin")
-        origin.push(env=ssh_env)
+        # A rejected push does not raise in GitPython; surface it explicitly.
+        origin.push(env=ssh_env).raise_if_error()
         for index, mirror_url in enumerate(mirrors):
             mirror = repo.create_remote(f"mirror_{index}", url=mirror_url)
-            mirror.push(env=ssh_env)
+            mirror.push(env=ssh_env).raise_if_error()
```
nikbpetrov: so if the push fails now, we will be doing `sys.exit(1)`

push, if they have not changed something upstream, fails if two simultaneous pushes happen (ask me how I know.... I am looking at you, Pavlovia)

that means that if we ever have simultaneous pushes, something will fail; quick inspection shows me that leaderboard is executing jobs in parallel and is pushing via the `clone_commit_and_push`

i'd recommend checking this push behaviour docs/code if uncertain

if correct, then i'd suggest a small retry/backoff -- that would actually solve the case where an intermittent github api unavailability can fail whole jobs (`sys.exit(1)` will indeed fail a whole job and leaderboard jobs are expensive)

src/nightly_update_workflow/manager/main.py
```diff
@@ -190,14 +190,28 @@ def main():
...
+    # one process clones and pushes to the dataset repo, avoiding a race condition.
+    dict_to_use_push_resolution_sets = "push_resolution_sets"
+    operation_push_resolution_sets = call_worker(
+        dict_to_use=dict_to_use_push_resolution_sets,
+        task_count=1,
+    )
+    cloud_run.block_and_check_job_result(
+        operation=operation_push_resolution_sets,
+        name=dict_to_use_push_resolution_sets,
+        exit_on_error=False,
```
nikbpetrov: no slack msg?

nikbpetrov: and not failing the job (`sys.exit(1)`) -- sohuldnt we want to know when pushing resolution sets failed

i think this will get rectified next day but still

src/orchestration/func_push_resolution_sets/requirements.txt
```diff
@@ -0,0 +1,13 @@
...
+termcolor
+slack_sdk
+pandera
+pytz
+python-dateutil
+backoff
```
nikbpetrov: is this used? probably should be re the other commit/comment to add a retry on git but worth checking deps

src/helpers/git.py
oskariluoma: Thanks for spotting this! My worry with retry/backoff is that waiting might not help in every case here. If I've understood it right, once another job has pushed, our clone is stale, so every retry sends the same push and gets the same rejection.

I was thinking a fetch+rebase onto origin/main before each retry might handle that, and it should still cover the intermittent GitHub case you mentioned, since the rebase does nothing when nothing has moved upstream. The three jobs also seem to write to different files (leaderboard_tournament.*, leaderboard_baseline.*, leaderboard_preliminary.*), so conflicts wouldn't come up in practice.

Does that match your experience, or is there something I'm missing that would make retry alone work? I was also wondering about splitting the mirror push into its own block so a Hugging Face failure logs an error instead of failing the whole job, since origin already has the data by then. But happy to leave that out if you'd rather keep it simple.

src/nightly_update_workflow/manager/main.py
oskariluoma: Yep! Setting `exit_on_error=True` makes sense. The slack msg and the `sys.exit(1)` are in the same branch in `block_and_check_job_result`, so `False` was suppressing the alert too.

One thing to confirm: this also makes it fatal for the rest of the run. The leaderboards and website are launched after this check, so a failed push means they never start. They don't read the resolution sets, so it's collateral damage. Fine with that, or would you rather I split the flag so we alert without exiting?

houtanb: Actually, you can/should run the git push at the end. You can launch the job after the website is created even.  then set `exit_on_error=True`. Put it after the question set creation checks. You can then push all resolution sets + leaderboards in one git commit as well, which would be ideal

oskariluoma: Done: [e318934] moved the push job to the end of main(), after the question set and forecaster checks, and set exit_on_error=True.

On bundling the leaderboards into the same commit, do you want that here or a separate PR?

houtanb: you can put that in the same PR. Thanks!

houtanb: Actually, @oskariluoma I take that back. If you haven't done it don't do it. leave as is. There's another issue where that would slot in well https://github.com/forecastingresearch/forecastbench/issues/181 and you can work on that next

src/orchestration/func_push_resolution_sets/requirements.txt
oskariluoma: I checked and at the moment backoff and slack_sdk could be removed. However I'm not doing anything yet because that could change regarding the git.py thread above. I copied these from func_resolve at the start and forgot to come back to check, thanks for pointing this out.

src/helpers/git.py
nikbpetrov: fetch+rebase works push if you do all of this in a single commit (as per below), that should be resolved

oskariluoma: Sorry, just want to make sure I read that right. Do you mean I should add the fetch+rebase, or that the single-commit approach already covers it so we can skip it?

oskariluoma:
I suggest merging the PR that solves #181 before this one. I'll have the PR up soon. It will bundle up the leaderboard pushes into a single commit. But at the moment the three leaderboard jobs run in parallel and their pushes can reject each other. The `raise_if_error()` added here would fail the nightly run.

houtanb:
Hmm, I don't see why. If that one goes after this one, you'd just add a few more files to push in the commit right? One overall commit containing resolution files, leaderboards, parity dates, and question difficulty per night? Instead of `func_push_resolution_sets` you'd rename it to be more general `func-push-datasets-to-git`. I think that should come after this or be folded into this if you want. Happy to discuss on a call or Slack, in case I'm missing something in what you see

src/helpers/git.py
```diff
@@ -84,50 +86,45 @@ def clone_and_push_files(
...
+    if not error_encountered:
+        for index, mirror_url in enumerate(mirrors):
+            try:
+                mirror = repo.create_remote(f"mirror_{index}", url=mirror_url)
+                mirror.push(env=ssh_env).raise_if_error()
+                repo.delete_remote(mirror.name)
+                logger.info(f"Pushed to {mirror_url} (mirror)")
+            except Exception as e:
+                message = e.message if hasattr(e, "message") else str(e)
+                logger.warning(f"Could not push to {mirror_url} (mirror): {message}")
```
nikbpetrov: love this handling of down or diverged but overall logging might be a tad futile as nobody reads these - maybe a slack msg w/ warning so that the 'alert' becomes fails on 2x nights (that a human must notice but at least we have a way) --  @houtanb do you monitor these loggers?

src/helpers/git.py
nikbpetrov: good catch by CC (checked)

Missing-secret crash (git.py:33). clone() returns None when API_GITHUB_SSH_ID_RSA isn't set, so the tuple unpack in clone_and_push_files raises TypeError instead of the intended soft skip. Pre-existing bug, not a regression — but this PR puts it on a must-succeed nightly path, so worth fixing while here (return early before cloning, as RUNNING_LOCALLY does).

nikbpetrov:
Fine PR, solves the problem it sets out but I catch myself asking if these really are problems we should be solving at all, @houtanb .

The whole GCP-tied architecture remains strange to me, tests are still lacking and I am nervous about every line. There's a ton of assumptions built on top of each other, and all I want to do is just point the code to some folder (on GCS or local) and have all the code just work from whatever state the files are in - each stage must encode exactly what it expects, and test must be the encoding of these expectations.

This is obviously not the case here as every stage is very finnicky, local runs/test etc are borderline unrunnable etc. We've talked about this before but I feel like we're doing a lot of patching these days. Not immediately actionable any of this I guess as such a refactor would be significant amount of work*. This codebase currently is worthwhile because it's been battle-tested a decent bit, which is impressive.

*although, tbh, as long as we get the tests right, agents can probably execute the majority of it and shadow parity runs/code review over the course of a few weeks can verify correctness/architecture --- even if minor bugs slip through after all that, at least will be debugging/patching something more robust.

houtanb:
@nikbpetrov indeed that's why you're working on the refactor and a test suite!

It takes time because we're doing this on a public-facing system so we're going slowly to be confident in changes, but the end goal, depending on tradeoffs, is a more reliable system that will allow us to move more quickly. All ideas are welcome.

src/helpers/git.py
houtanb: I only check them when a job breaks. Sending Slack messages is preferred

oskariluoma: Thanks, good to know! I left the logs in place and added a slack msg if the mirror push fails.

houtanb:
merged in cffaab6

PR #250, opened 2026-07-02
fix(metadata): give gpt-5-mini reasoning headroom for tagging/validation succeed

### LINKED ISSUE #249: Metadata tagging & validation silently mislabel all newly-processed questions (gpt-5-mini reasoning starves the token   budget)
- After the “rewrite LLM forecaster” refactor (`6566923`, 2026-06-26), metadata jobs call `gpt-5-mini` with low `max_output_tokens` and no `reasoning.effort`. Since the model defaults to medium reasoning, every request exhausts the budget and fails with `status=incomplete, reason=max_output_tokens`. After retries, `tag_questions` / `validate_questions` fall back to `category="Other"` and `valid_question=True`, silently mislabeling new questions.

- A Kalshi tagging run made 120 requests; 0 succeeded, then hit the 30-minute timeout.

- Questions processed through this path since 2026-06-26 may be contaminated.

- Fix: Set `reasoning={"effort": "minimal"}` and raise the tagging/validation output budgets.

drozado:
Fixes #249 
  
  `gpt-5-mini` is a reasoning model. Calling it with `max_output_tokens=50` (tag) / `500`
  (validate) and default (`medium`) reasoning makes it exhaust the budget on reasoning tokens and
  return `status=incomplete (max_output_tokens)` every time, so metadata silently fell back to
  `category="Other"` / `valid_question=True`.
  
  ## Changes
  - `helpers/metadata_llm.py`: pass `reasoning={"effort": "minimal"}` (covers tag + validate).
  - `tag_questions/main.py`: `max_output_tokens` 50 → 512.
  - `validate_questions/main.py`: `max_output_tokens` 500 → 1024.
  - `test_model_request_params.py`: assert a reasoning policy is *present* rather than pinning the
    exact effort value, so future model/reasoning/budget changes don't break the test.

## Summary by CodeRabbit

* **New Features**
  * Improved metadata responses with lighter reasoning guidance for faster, more efficient results.
  * Increased response limits for question tagging and validation to reduce the likelihood of truncated outputs.

* **Bug Fixes**
  * More reliable question categorization and validation when generating longer model responses.

* **Tests**
  * Updated automated checks to reflect the new model response settings.

### DIFF
--- a/src/tests/test_model_request_params.py
+++ b/src/tests/test_model_request_params.py
@@ -138,6 +138,7 @@ def test_metadata_model_response_routes_through_shared_model_run(monkeypatch):
             "Classify this question.",
             {
                 "max_output_tokens": 123,
+                "reasoning": {"effort": "minimal"},
                 "safety_identifier": "forecastbench-safety-id",
             },
         ),
--- a/src/helpers/metadata_llm.py
+++ b/src/helpers/metadata_llm.py
@@ -32,5 +32,6 @@ def get_metadata_model_response(prompt: str, max_output_tokens: int) -> str:
     return _get_metadata_model_run().get_response(
         prompt=prompt,
         max_output_tokens=max_output_tokens,
+        reasoning={"effort": "minimal"},
         safety_identifier=get_openai_safety_identifier(),
     )
--- a/src/metadata/tag_questions/main.py
+++ b/src/metadata/tag_questions/main.py
@@ -34,7 +34,7 @@ async def _get_category_single(index, row, semaphore):
             response = await asyncio.to_thread(
                 metadata_llm.get_metadata_model_response,
                 prompt=prompt,
-                max_output_tokens=50,
+                max_output_tokens=512,
             )
             category = response.strip('"').strip("'").strip(" ").strip(".")
             logger.info(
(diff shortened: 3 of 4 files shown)


src/tests/test_model_request_params.py
```diff
@@ -131,6 +131,8 @@ def fake_get_model_run(model_run_key):
     )
 
     assert response == "metadata response"
+    assert "reasoning" in calls[1][2]  # reasoning policy is set (any value)
```
nikbpetrov: maybe set this test to check that `reasoning` is exactly `{"effort": "minimal"}`; also helps catch typos

nikbpetrov:
@houtanb This is a good catch for a bug and raises the possibility that not all LLM args/setup is being forwarded/works exactly as expected. Perhaps a smoke test-like cli can run a few more reqests and check outputs for systematic problems/refusals. But that's separate.

LGTM

src/tests/test_model_request_params.py
drozado: Good point. Updated the test to assert the exact reasoning policy for the currently pinned metadata model. If/when we migrate this helper to a newer model, this test should fail and force us to revisit the model-specific reasoning setting as newer models use "effort": "none".

Issue #257, opened 2026-07-23
add logo for Hunchfox


houtanb:
<img width="800" height="800" alt="Image" src="https://github.com/user-attachments/assets/b1119148-2cce-49ef-8f09-925da829ad35" />

jmacd867:
I'll take it on!

Issue #258, opened 2026-07-23
add logo for Knowable


houtanb:
<img width="632" height="632" alt="Image" src="https://github.com/user-attachments/assets/459b6756-3ef0-4ced-90a8-76cffd62eadc" />

jmacd867:
Will also do this

Issue #256, opened 2026-07-23
add logo for Vaticinus


houtanb:
<img width="512" height="512" alt="Image" src="https://github.com/user-attachments/assets/615b24ea-66bd-45f1-9564-1a135daa8ad4" />

jmacd867:
Will also take this on

PR #259, opened 2026-07-23
logos added,

jmacd867:
note that they aren't added to the assets because of the .gitignore


jmacd867:
Closes #257 #256 #258

PR #291, opened 2026-09-04
feat: expand DBnomics questions with curated ECB series

pythoryn:
## Summary

Adds 458 curated ECB series to DBnomics, expanding it from 53 to 511 series.

Increases the DBnomics job memory and timeout limits to support the larger workload.

Excludes series with missing data or no non-missing observation within 10 days from new question sets and sends a Slack notification. Series become eligible again when fresh observations resume.

### DIFF
--- a/src/sources/dbnomics.py
+++ b/src/sources/dbnomics.py
@@ -50,7 +50,10 @@ _QUESTION_TEMPLATES = {
 }
 
 _VALUE_EXPLANATIONS = {
-    "meteofrance": "The daily average temperature at the French weather station at {station}.",
+    "meteofrance": (
+        "The latest value published as of the question-set freeze date for the daily average "
+        "temperature at the French weather station at {station}."
+    ),
     "ecb": (
         "The latest value published as of the question-set freeze date for "
         "“{question_subject}” in the European Central Bank’s “{dataset_name}” dataset."
--- a/src/sources/dbnomics.py
+++ b/src/sources/dbnomics.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from datetime import date
 from typing import Any, ClassVar
 
 import backoff
@@ -16,10 +17,15 @@ from _schemas import DbnomicsFetchFrame, QuestionFrame, ResolutionFrame
 from helpers import constants, data_utils, dates
 
 from ._dataset import DatasetSource
+from .dbnomics_questions import ECB_QUESTIONS, METEOFRANCE_STATIONS
 
 logger = logging.getLogger(__name__)
 
 _BASE_URL = "https://api.db.nomics.world/v22/series/"
+_PROVIDER_NAMES = {
+    "ECB": "European Central Bank",
+    "meteofrance": "Météo-France",
+}
 
 # Some dataseries with regular updates have large numbers of NA values during periods in which
 # data is not being reported. _OBSERVATIONS_WITHOUT_DATA detects these quiet periods and excludes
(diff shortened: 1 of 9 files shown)


src/sources/dbnomics.py
houtanb: this can now just be `dfq.loc[dfq["id"] == id, "freeze_datetime_value"] = freeze_datetime_value`

src/sources/dbnomics.py
```diff
@@ -189,17 +188,9 @@ def update(
...
             series_values = df_series["value"]
-            series_dates = pd.to_datetime(df_series["period"])
-
-            last_fetch_date = series_dates.iloc[-1]
-            last_fetch_value = series_values.iloc[-1]
-            freeze_datetime_value = (
-                float(last_fetch_value)
-                if last_fetch_date.date() > yesterday and last_fetch_value != "NA"
-                else "N/A"
-            )
```
houtanb: It looks like you got rid of this. But, I think for daily data like meteofrance, if the value is not present, then the freeze value shoud be marked as `"N/A"`. Imagine weather was not reported for a month in some location. It would be incorrect to just draw forward last month's weather value either for comparison or for resolution

src/sources/dbnomics.py
houtanb, line 231: Just confirming all these series are updated daily or weekly, correct? If not, this needs to be adjusted top drop the first forecast horizon

houtanb:
@pythoryn I'd also drop the dbnomics API key given we can call without it and there's no documentation about how to use it. We can add it back once we need it or at least it has been documented

houtanb:
@pythoryn is there any notification of series that fail quietly, e.g. one that's present but no longer updated? Perhaps some slack message to look into series that haven't been updated in, say, 7 days?

src/helpers/data_utils.py
```diff
@@ -54,7 +54,17 @@ def download_and_read(filename, local_filename, df_tmp, dtype):
         filename=filename,
         local_filename=local_filename,
     )
-    df = pd.read_json(local_filename, lines=True, dtype=dtype, convert_dates=False)
+    if os.path.getsize(local_filename) == 0:
+        return df_tmp
+    # Read large JSONL files in chunks to avoid a temporary whole-file memory spike.
```
nikbpetrov: comment is unnecessary; instead can you either open an issue or sth to centralize that chunked file reading inside _source_io or data_utils or sth so taht all jsons are read chunked (or perhaps certain ones that match some condition -- maybe there is a ready-made implementaiton of that logic though whoever addresses that issue in the future should decide)

src/orchestration/_source_io.py
```diff
@@ -25,8 +27,11 @@ def write_fetch_output(source: str, dff: pd.DataFrame) -> None:
     filenames = data_utils.generate_filenames(source)
     local = filenames["local_fetch"]
     with open(local, "w", encoding="utf-8") as f:
-        for record in dff.to_dict(orient="records"):
-            f.write(json.dumps(record, ensure_ascii=False) + "\n")
+        for start in range(0, len(dff), _FETCH_WRITE_CHUNK_SIZE):
+            # Process a small batch at a time to avoid running out of memory.
```
nikbpetrov: ditto as above, though according to CC and its tests, chunked writes dont do much; in any case new issue regarding this and a centralized reading/writing mechanism will be hte way to go

nikbpetrov:
Okay, interesting finding: `/tmp` on GCP is [memory-backed](https://docs.cloud.google.com/run/docs/container-contract#filesystem) (hence the OOM that the resource bump is solving). I asked CC to run some tests: the update procedure today uses just under 1 GiB memory all-in today. If we double the file size, which will eventually happen, we still fit in the 2 GiB memory limit set, but then it will fail. I guess we can just address then?! For now, that might suffice, but I'd rather a more general solution that monitors these resources and sends a slack message if we are 80% there so we can be proactive about it.

I don't know how that looks exactly. Maybe GCP already has some built-in solution for that, or maybe we set up some centralized monitor that runs over all jobs, or maybe this becomes part of my/our testsuite that we do for every merge.

@houtanb will let you decide how to handle this - an issue?

src/tests/orchestration/test_source_io.py
```diff
@@ -0,0 +1,40 @@
...
+
+import json
+
+import pandas as pd
+
+from helpers import env
+from orchestration import _source_io
+
+
+def test_write_fetch_output_preserves_jsonl_rows_and_uploads(monkeypatch, tmp_path):
```
nikbpetrov:
```python
def test_fetch_output_round_trips_through_download_and_read(monkeypatch, tmp_path):
    n = 50_010  # crosses both the 10k write chunk and the 50k read chunk
    dff = pd.DataFrame({
        "id": [f"s{i}" for i in range(n)],
        "period": ["2026-01-01"] * n,
        "value": [1.5] * (n - 3) + ["NA", 2.0, "NA"],   # mixed only in the last chunk
        "series_name": ["Météo-France"] * n,
    })
    path = tmp_path / "x_fetch.jsonl"
    monkeypatch.setattr(_source_io.data_utils, "generate_filenames",
                        lambda s: {"local_fetch": path, "jsonl_fetch": "x_fetch.jsonl"})
    monkeypatch.setattr(_source_io.gcp.storage, "upload", lambda **k: None)
    monkeypatch.setattr(data_utils.gcp.storage, "download_no_error_message_on_404", lambda **k: None)

    _source_io.write_fetch_output("x", dff)
    out = data_utils.download_and_read("x_fetch.jsonl", path, pd.DataFrame(), {"id": str})

    pd.testing.assert_frame_equal(out, dff)
```

nikbpetrov: though tests are in a pretty bad shape everywhere, so dont' sweat it - this one just caught my eye

houtanb:
For the memory change, I like the idea of this being something more general, but until we're at the memory limit (32GB or 64GB, don't remember) don't think it's worth implementing. I think this was a good initiative but think we have bigger fish to fry.

There are changes on the horizon to the way we handle the question bank: either move to a DB or store all resolutions in one file. Whichever we choose will change what we do here anyway.

So, at the moment, I'm for removing it altogether, increasing the memory limit, and moving on.

pythoryn:
Thanks for the review. I’ve removed the chunked JSONL changes and DBnomics API key, increased the job memory and timeout limits, and reused the calculated freeze value as suggested.

All 458 added ECB series have daily or business-daily observation frequency, so I’ve retained the one-week forecast horizon. Series with no fetched data, or whose latest non-missing observation is more than 10 days old, are excluded from new question sets and flagged in a Slack notification. They become eligible again when fresh observations resume.

I have updated the PR description, and CI checks are passing. fetch() and update() run without issues in my GCP sandbox. Please let me know if anything else is needed before merging.

src/sources/dbnomics.py
```diff
@@ -16,132 +17,133 @@
...
     "meteofrance": (
         "What is the probability that the daily average temperature at the French weather station "
         "at {station} will be higher on {resolution_date} than on {forecast_due_date}?"
-    )
+    ),
+    "ecb": (
+        "What is the probability that the value of the latest observation dated on or before "
+        "{resolution_date} for the European Central Bank time series “{question_subject}” "
+        "will exceed the value of its latest observation dated on or before "
+        "{forecast_due_date}?"
```
houtanb, line 48: "on or before" means they have to consider whether or not the value will be updated? Why not just "on" the forecast due date? If there's an "N/A" value on that date, the question is nullified (I believe that's what we do with meteofrance, please verify this).

We'd then need to confirm sure nan is being stored in the resolution values (not forward filled)

src/sources/dbnomics.py
```diff
@@ -16,132 +17,133 @@
...
+    ),
 }
 
 _VALUE_EXPLANATIONS = {
-    "meteofrance": "The daily average temperature at the French weather station at {station}."
+    "meteofrance": "The daily average temperature at the French weather station at {station}.",
+    "ecb": (
+        "The latest value published on or before the question-set freeze date for "
+        "“{question_subject}” in the European Central Bank’s “{dataset_name}” dataset."
+    ),
```
houtanb, line 60: this can be "The latest value published as of the question-set freeze date..."

Perhaps meteofrance should be changed to "The latest published daily average temp..." (in a separate commit)

src/sources/dbnomics.py
pythoryn: Hi Houtan, I verified that meteo France preserves missing values and excludes forecasts from scoring when either comparison date has no value. We can apply the same approach to ECB, but our forecast due dates fall on Sundays, and 213 of the proposed ECB series report only on business days (i.e. their cadence is business-daily, that is, Monday to Friday). Without carrying earlier values forward, those series would generally have no Sunday baseline, so none of their forecast horizons could be scored. That’s the reason for “on or before.” We could use separate wording and handling for business-day series, though that would add some complexity.

src/sources/dbnomics.py
pythoryn: Agreed, “The latest value published as of the question-set freeze date…” is more clear. I’ll update the ECB wording accordingly.

Regarding meteofrance: I'm ok with your suggestion. I’ll update the meteo France wording in a separate commit.

PR #293, opened 2026-09-05
Fix/287 yfinance delisted renamed

### LINKED ISSUE #287: yfinance delisted / renamed stocks
Logging error:
> yfinance fetch: 5 ticker(s) failed to fetch and are no longer in the S&P 500 (likely delisted or renamed): CTRA, MTCH, IPG, HOLX, SATS. If delisted, add to nullified_questions; if renamed, add to ticker_renames (mapping to the replacement symbol).

Look into each stock, choose the corrected path (nullified + nullified date / rename)

oskariluoma:
fix: update delisted and renamed yfinance tickers

IPG was delisted 2025-11-28 when the Omnicom acquisition closed: https://www.omc.com/newsroom/omnicom-completes-acquisition-of-interpublic-forming-the-worlds-leading-marketing-and-sales-company-built-for-intelligent-growth-in-the-next-era/ ; https://stockanalysis.com/stocks/ipg/

HOLX was delisted 2026-04-07, taken private by Blackstone and TPG: https://www.hologic.com/about/press-release/blackstone-and-tpg-complete-acquisition-hologic ; https://stockanalysis.com/stocks/holx/

CTRA was delisted 2026-05-07, merged into Devon Energy: https://investors.devonenergy.com/investors/press-releases/press-release-details/2026/Devon-Energy-and-Coterra-Energy-Complete-Merger/ ; https://stockanalysis.com/stocks/ctra/

SATS was renamed to ECHO on 2026-06-24, still trading on Nasdaq: https://www.lightreading.com/satellite/echostar-to-change-stock-ticker-to-echo- 

MTCH is neither delisted nor renamed, still trading on Nasdaq: https://stockanalysis.com/stocks/mtch/ It only dropped out of the S&P 500 in the March 2026 rebalance, and the fetch failure had a different cause. For some reason Yahoo did not return "longName" value for MTCH which caused a delisting. Added a fall back to shortName to restore the normal fetch. 

Nullification dates are last trading day + 1, so 2025-11-27 for IPG, 2026-04-07 for HOLX and 2026-05-07 for CTRA. Verified those against the last real price change in each resolution file.

Also added a test to see if the MTCH handling worked, let me know if you think this isn't needed. 

Closes #287

### DIFF
--- a/src/sources/yfinance.py
+++ b/src/sources/yfinance.py
@@ -146,20 +146,18 @@ class YfinanceSource(DatasetSource):
                     }
                 )
                 logger.info(company_name)
-            elif (
-                company_name is None
-                and ticker_symbol in set_current
-                and ticker_symbol not in set_top_500
-            ):
-                # In the question pool, no longer fetchable, and out of the S&P 500, but not in any
-                # curated list: either newly delisted or newly renamed. Carry the existing row
-                # forward as resolved and record it so the driver can flag it for triage.
+            elif ticker_symbol in set_current and ticker_symbol not in set_top_500:
+                # In the question pool, no usable price data (no name or no price history), and
+                # out of the S&P 500, but not in any curated list: either newly delisted or newly
+                # renamed. Carry the existing row forward as resolved and record it so the driver
+                # can flag it for triage.
                 rows.append(self._carry_forward_resolved(ticker_symbol, dfq, current_time))
                 self.uncurated_delisted_tickers.append(ticker_symbol)
                 logger.warning(
-                    f"{ticker_symbol} failed to fetch and is no longer in the S&P 500 (likely "
-                    "delisted or renamed). If delisted, add it to nullified_questions; if renamed, "
-                    "add it to ticker_renames (mapping it to its replacement symbol)."
+                    f"{ticker_symbol} returned no usable price data (no name or no price "
+                    "history) and is no longer in the S&P 500 (likely delisted or renamed). "
+                    "If delisted, add it to nullified_questions; if renamed, add it to "
+                    "ticker_renames (mapping it to its replacement symbol)."
                 )
 
         return pd.DataFrame(rows)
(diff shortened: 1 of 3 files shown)


houtanb:
@oskariluoma please close the issue from the commit message: https://github.com/forecastingresearch/forecastbench/issues/287.

Please add the above info to the commit message body as well: https://github.com/forecastingresearch/forecastbench/pull/293#issue-5356522690

please change the commit message to start with `fix: update delisted and renamed yfinance tickers"

src/sources/yfinance.py
```diff
@@ -306,11 +306,13 @@ def _fetch_one_stock(self, ticker_symbol: str) -> tuple[str | None, pd.DataFrame
...
         Returns:
-            Tuple of (company_name, hist_df) or (None, None) on failure.
+            Tuple of (company_name, hist_df), where company_name is None when yfinance has
+            no name for the ticker, or (None, None) on failure.
         """
         try:
             ticker = yf.Ticker(ticker_symbol)
-            company_name = ticker.info["longName"]
+            info = ticker.info
+            company_name = info.get("longName") or info.get("shortName")
```
nikbpetrov, line 313: This is a bit of a change (positive one most likely), but I am trying to work out the implications. So we used to treat a no-`longName` stock as delisted/renamed and was flagged for triage in `fetch()`. Now, if we have a `shortName` (with no history) for a question already in the bank, then neither of the `if` branches in `fetch()` trigger so the question remains stale? I guess a quick check on whether that's possible now, or in the future, would be helpful. Does that make sense?

nikbpetrov:
1 comment above, otherwise LGTM (love the comments/explanations in the PR so I dont have to rediscover those myself).

src/sources/yfinance.py
oskariluoma: Yeah, makes sense. With a name but no history neither branch fires.

Did a quick check and it doesn't look possible right now: for a delisted/renamed company
Yahoo seems to return every value field as None. Which should lead to every delisting/renaming hitting the branch. But it's definitely possible, e.g. maybe a long trading halt or just Yahoo not returning values. 
I think this could be solution: Making the elif so it doesn't depend on which name fields Yahoo happens to return:

src/sources/yfinance 149:152:
```
    elif (
        (company_name is None or hist.empty)   #adding the 'or hist.empty'
        and ticker_symbol in set_current
        and ticker_symbol not in set_top_500
    ):
```

So now this would also catch the hist.empty scenario and solve this problem? Let me know if I misunderstood something.

nikbpetrov: Makes sense to me, that `logger.warning` msg would also need a tiny update too in this case. @houtanb just flagging for a second pair of eyes just in case.

houtanb: @oskariluoma @nikbpetrov that looks good to me, small change:
```python
elif ticker_symbol in set_current and ticker_symbol not in set_top_500:
```
should do the job since `company_name is None or hist.empty` is implied by the failed condition on the `if` above

PR #297, opened 2026-09-11
fix: clarify Dbnomics ECB question wording and missing-observation fallback

pythoryn:
Simplifies the ECB question template by stating the comparison first and explaining the missing-observation fallback in a separate sentence. Resolution behavior is unchanged.

Updates the corresponding test expectation. Follow-up to #291 and Houtan’s feedback.


houtanb:
@pythoryn do I need to delete the `dbnomics_questions.jsonl` file before this will take effect?

houtanb:
merged d0cbe8d5c instead

